PR #26, opened 2024-07-29
Additional client_registration_types MAY be defined and used

selfissued:
Added a clarification suggested by @jricher in his role as Designated Expert for the IANA OAuth Dynamic Client Registration Metadata registry.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -1273,6 +1273,7 @@
               are
               <spanx style="verb">automatic</spanx>
               and <spanx style="verb">explicit</spanx>.
+	      Additional values MAY be defined and used.
             </t>
           </list>
         </t>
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -1273,6 +1273,7 @@
               are
               <spanx style="verb">automatic</spanx>
               and <spanx style="verb">explicit</spanx>.
+	      Additional values MAY be defined and used.
```
jricher, line 1276:
```suggestion
	      Additional values MAY be defined and used, without restriction by this specification.
```

openid-federation-1_0.xml
```diff
@@ -1351,6 +1352,7 @@
               Federation-type values defined by this specification are
               <spanx style="verb">automatic</spanx>
               and <spanx style="verb">explicit</spanx>.
+	      Additional values MAY be defined and used.
```
jricher, line 1355:
```suggestion
	      Additional values MAY be defined and used, without restriction by this specification.
```

jricher (commented):
I would recommend adding the proposed caveat to the text, if that is amenable. The intent is to say that the values defined are unrestricted, and so an implementation will need to deal with that and not just crash if there's an unknown.

rohe (approved):

peppelinux (approved):

PR #27, opened 2024-07-30
Entity Identifiers use the https scheme

### LINKED ISSUE #22: Entity Identifier URI scheme not defined
Nowhere in the specification do we directly say that an Entity Identifier must use the `https` scheme.

I'm thinking that the best place to do this is in the Entity Identifier definition itself.  Thoughts?

selfissued:
Fixes #22

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -230,7 +230,10 @@
             </t>
             <t hangText="Entity Identifier">
               <vspace/>
-              A globally unique URI that is bound to one Entity.
+              A globally unique URL that is bound to one Entity.
+	      This URL MUST use the <spanx style="verb">https</spanx>
+	      scheme and MAY contain port and path components.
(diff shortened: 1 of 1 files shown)


rohe (approved):

openid-federation-1_0.xml
```diff
@@ -232,7 +232,10 @@
             </t>
             <t hangText="Entity Identifier">
               <vspace/>
-              A globally unique URI that is bound to one Entity.
+              A globally unique URL that is bound to one Entity.
```
adeinega, line 235: Usually, a fragment identifier (component) is simply forbidden.

peppelinux (approved):

openid-federation-1_0.xml
selfissued: The current language forbids it:
> It MAY NOT contain query parameter or fragment components.

Issue #30, opened 2024-08-06
Ambiguity in `fetch` endpoint response when `iss` and `sub` are the same


jcmelati:
Hello!

I believe I have identified an inconsistency in the OpenID Federation specification related to the `/fetch` endpoint, specifically when the `iss` and `sub` parameters are the same. In this case, the endpoint currently returns an entity configuration, in accordance with [section 8.1.1](https://openid.net/specs/openid-federation-1_0.html#section-8.1.1-2.4) which states regarding the `sub` attribute that "_if this parameter is omitted, it is considered to be the same as the issuer and indicates a request for a self-signed **Entity Configuration**_".

The specification also states that "_an Entity **MUST publish Subordinate Statements** about its Immediate Subordinates **via its fetch endpoint**_" ([source](https://openid.net/specs/openid-federation-1_0.html#section-8.1-1)).

However, this is problematic because:

- The `fetch` endpoint is intended to retrieve `Subordinate Statements`.
- Although `Entity Configuration Statement` and `Subordinate Statement` have similar data structures, they are not exactly the same.
- An entity cannot logically be a subordinate of itself.

Proposal:
To resolve this, the specification should be updated so that when `iss` and `sub` are the same in a `/fetch` request, the endpoint returns [invalid_request (Bad Request 400)](https://openid.net/specs/openid-federation-1_0.html#section-8.9-2.2.2.2). This will clarify that the `fetch` endpoint should only be used to retrieve Subordinate Statements, not Entity Configurations.

Supporting evidence:
In the Italian federation, we've observed that the payloads from the `/.well-known/openid-federation` and `/fetch` endpoints differ significantly, specifically on the metadata part that is present on the Entity Configuration Statement, but not on the Subordinate Statement. 

Suggested Action:
Amend the spec to explicitly state that the fetch endpoint should return `invalid_request` (400 Bad Request error) if `iss` and `sub` are identical.

selfissued:
We discussed this on the 7-Aug-2 Federation editor's call.  We agree on returning `invalid_request` when they are equal as you suggest.

selfissued:
I agree that the spec is currently self-contradictory.  The fetch endpoint describes that it can be used to retrieve either ECs or SS.  It's described in terms of Entity Statements - not SSs.  But there's the statement you cited that says it's about SSs:
[...]

PR #31, opened 2024-08-07
fix: the parameters sub and iss must not be set with the same value at the fetch endpoint

peppelinux:
This PR resolves https://github.com/openid/federation/issues/30


selfissued:
We need to decide whether to remove the ability to retrieve ECs with Fetch before considering this PR (which removes it).

See the question at https://github.com/openid/federation/issues/30#issuecomment-2274464616.

selfissued (changes requested):
I believe this can be closed without being merged since it is replaced by #38 .

selfissued:
Replaced by #38

PR #38, opened 2024-08-13
Simplify fetch endpoint to only return Subordinate Statements

### LINKED ISSUE #30: Ambiguity in `fetch` endpoint response when `iss` and `sub` are the same

selfissued:
Fixes #30 
Replaces #31
Cc: @jcmelati

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -24,7 +24,7 @@
 
   <front>
     <title abbrev="OpenID Federation">OpenID Federation 1.0 -
-      draft 37
+      draft 38
     </title>
 
     <author fullname="Roland Hedberg" initials="R." role="editor"
@@ -87,7 +87,7 @@
       </address>
     </author>
 
-    <date day="29" month="July" year="2024"/>
+    <date day="12" month="August" year="2024"/>
 
     <workgroup>OpenID Connect Working Group</workgroup>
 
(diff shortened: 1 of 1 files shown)


vdzhuvinov (approved):

ve7jtb (approved):
LGTM

PR #51, opened 2024-08-23
Improve descriptions of JWT claims being registered

selfissued:
Per feedback from the IANA Designated Expert.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -598,9 +598,9 @@
               OPTIONAL.
 	      Entity Statements require that
 	      the <spanx style="verb">crit</spanx> (critical) claim
-	      defined in <xref target="critClaim"/> be understand and processed.
+	      defined in <xref target="critClaim"/> be understood and processed.
 	      Claims in the Entity Statement whose claim names are in the array
-	      that is the value of this claim MUST be understand and processed.
+	      that is the value of this claim MUST be understood and processed.
 	      Claims specified for use in Entity Statements by this specification
 	      MUST NOT be included in the list.
             </t>
@@ -6739,7 +6739,7 @@ HTTP/1.1 302 Found
 	<t>
 	  This claim is used in this specification in <xref target="entity-statement"/>
 	  to identify claims not defined by this specification
-	  that MUST be understand and processed
+	  that MUST be understood and processed
 	  when used in Entity Statements.
 	</t>
       </section>
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -3261,35 +3261,6 @@
           cross-JWT confusion, per Section 3.11 of <xref target="RFC8725"/>.
         </t>
 
-        <section title="Trust Mark Delegation" anchor="trust_mark_delegation">
-          <t>
-            There will be cases where the owner of a Trust Mark for some reason
-            may not match the Trust Mark issuer due to administrative or
-            technical requirements. Take as an example vehicle
-            inspection. Vehicle inspection is a procedure mandated by national
-            or subnational governments in many countries, in which a vehicle is
-            inspected to ensure that it conforms to regulations governing
-            safety, emissions, or both. The body that mandates the inspections
-            does not perform them; instead, there may be commercial companies
-            performing the inspections, after which they issue the Trust Mark.
-          </t>
-          <t>
-            The fact that a Trust Mark is issued by a Trust Mark issuer that is
-            not the owner of the Trust Mark is expressed by including
-	    a <spanx style="verb">delegation</spanx> claim in the Trust Mark,
-	    whose value is a signed JWT.
-            The format of the delegation JWT is described below.
-          </t>
-          <t>
-            If the Federation Operator knows that Trust Marks with a
-            certain Trust Mark identifier may legitimately be issued by Trust Mark issuers
-            that are not the owner of the Trust Mark identifier, then information
-            about the owner and the Trust Mark identifier MUST be included in the
-            <spanx style="verb">trust_mark_owners</spanx> claim in
-	    the Trust Anchor's Entity Configuration.
-          </t>
-        </section>
-
         <section title="Trust Mark Claims" anchor="trust_mark_claims">
           <t>
 	    The claims in a Trust Mark are:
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -6733,42 +6736,72 @@ HTTP/1.1 302 Found
 	  as the <spanx style="verb">crit</spanx> value.
 	  Use of this claim is OPTIONAL.
 	</t>
+	<t>
+	  This claim is used in this specification in <xref target="entity-statement"/>
```
peppelinux, line 6740: claim-> Claim

openid-federation-1_0.xml
```diff
@@ -599,6 +599,8 @@
 	      Entity Statements require that
 	      the <spanx style="verb">crit</spanx> (critical) claim
 	      defined in <xref target="critClaim"/> be understand and processed.
+	      Claims in the Entity Statement whose claim names are in the array
+	      that is the value of this claim MUST be understand and processed.
```
rohe, line 603: Should not "understand" be "understood" ?

openid-federation-1_0.xml
```diff
@@ -6733,42 +6736,72 @@ HTTP/1.1 302 Found
 	  as the <spanx style="verb">crit</spanx> value.
 	  Use of this claim is OPTIONAL.
 	</t>
+	<t>
+	  This claim is used in this specification in <xref target="entity-statement"/>
+	  to identify claims not defined by this specification
+	  that MUST be understand and processed
```
rohe, line 6742: Again would not 'understood' be better ?

openid-federation-1_0.xml
selfissued: We're usually using "claim" in running text.

openid-federation-1_0.xml
selfissued: Thanks, fixed!

openid-federation-1_0.xml
selfissued: Thanks, fixed!

selfissued:
@bc_pi please review.

peppelinux (approved):

rohe (approved):

bc-pi (changes requested):
Definitely moving in the right direction, thank you, but some questions remain for me.

openid-federation-1_0.xml
```diff
@@ -6733,42 +6736,72 @@ HTTP/1.1 302 Found
...
+	  This claim is used in this specification in <xref target="entity-statement"/>
+	  to identify claims not defined by this specification
+	  that MUST be understood and processed
+	  when used in Entity Statements.
+	</t>
       </section>
 
       <section anchor="idClaim" title='"id" (Identifier) Claim'>
 	<t>
 	  The <spanx style="verb">id</spanx> (identifier) claim
```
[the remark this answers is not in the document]
selfissued: "id" as used by JSON-LD is a string.  So it's not in conflict.

This is half off-topic, but it's odd to me that JSON-LD didn't register any of its claims.  Part of my thinking was that it would be better get the "id" claim registered in a way that matches common usage than to leave it in limbo.  Then JSON-LD's usage would be conforming and could be considered a profile of the registered claim.

openid-federation-1_0.xml
```diff
@@ -7746,7 +7779,7 @@ HTTP/1.1 302 Found
 
       <section anchor="ClaimsRegistry" title="JSON Web Token Claims Registration">
 	<t>
-	  This specification registers the following claim in
+	  This specification registers the following claims in
```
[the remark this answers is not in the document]
selfissued: You raise a really good point.  We tried to register the claims used by JWTs defined by the specification that were currently unregistered.  But arguably the most important JWT defined by the spec - the Entity Statement - doesn't have all of its claims registered.  Thanks for pointing out what should have been obvious!

A quick skim makes me believe that these are such claims:  "jwks", "authority_hints", "metadata", "metadata_policy", "constraints", "metadata_policy_crit", "trust_marks", "trust_mark_issuers", "trust_mark_owners", "source_endpoint".

Of these, clearly "jwks", "metadata", and "constraints" should have broadly-usable definitions appearing in the list at https://openid.net/specs/openid-federation-1_0.html#name-general-purpose-jwt-claims.  My instinct for the rest is to register them as used by OpenID Federation but not generically.  This would parallel the way that "sub_jwk" is registered.

Do you agree with that course of action, Brian (or at least agree with it enough to review a PR on the topic)?  Others, please weigh in too!

openid-federation-1_0.xml
```diff
@@ -6733,42 +6736,72 @@ HTTP/1.1 302 Found
...
       </section>
 
       <section anchor="delegationClaim" title='"delegation" (Delegation) Claim'>
 	<t>
-	  The <spanx style="verb">delegation</spanx> claim
-	  expresses a delegation relationship.
-	  Like the <spanx style="verb">sub</spanx> (subject) claim,
-	  the processing of this claim is generally application specific.
+	  The <spanx style="verb">delegation</spanx> claim expresses that
+	  authority is being delegated to the party referenced in the claim value.
```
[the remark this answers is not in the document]
selfissued: Reading the referenced sections with fresh eyes, at a minimum, I'm seeing that the relationships between them need to be made clearer.  For instance, in the sentence:
> The format of the delegation JWT is described below.
"below" should be a section reference.  There are several instances of that kind of thing.

Likewise, in the _remarkable sentence_, there's no reference to the definition of the signed JWT in question.

I now plan to review all the text related to delegation and see if I can make it easier to follow and clearer.

You're also right that an example Delegation JWT is also in order.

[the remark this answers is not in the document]
selfissued: OK - I've reworked the delegation description to be much cleaner.  There's now a straightforward reference chain from the [delegation claim definition](https://openid.github.io/federation/mbj-jwt-claims.html#name-delegation-delegation-claim) to [its use in Trust Marks](https://openid.github.io/federation/mbj-jwt-claims.html#trust_mark_claims) to the definition of its value - the [Trust Mark delegation JWT](https://openid.github.io/federation/mbj-jwt-claims.html#delegation_jwt).

And the describing sentences are less remarkable as well. ;-)

Finally, #61 asks for creation of a Trust Mark delegation JWT example.

openid-federation-1_0.xml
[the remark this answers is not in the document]
selfissued: OK, the [latest commit](https://github.com/openid/federation/pull/51/commits/fa8a94a23e875c31537a546e37c2deaa046f297e) added registrations for the previously unregistered claims.  I shuffled the order of some of the claim registrations so that the referenced section numbers now occur in ascending order.

I believe that means that we're ready for you to re-review, @bc-pi.  Thanks!

PR #72, opened 2024-09-07
Allow multiple type request parameters in resolve requests

### LINKED ISSUE #43: Add return of federation_entity metadata if type is provided in resolve requests
This has been discussed before with no resolution. I just implemented this and found that I needed to encode some solution.

Section 5.2.2. states:

> It is RECOMMENDED that, when present, these metadata parameters occur in an Entity's federation_entity metadata. 
> They MAY also be present in the Entity's metadata for other Entity Types, particularly when the values for those Entity Types 
> differ from those for the federation_entity metadata. 

This means that metadata for any type may find vital information about that entity, not in the metadata of the selected metadata entity type, but in the federation_entity metadata.

This is a problem when a resolve request contains a `type` parameter. 

In the definition of resolve response in section 8.3.2, the content of the metadata parameter is defined as:

```
metadata
      REQUIRED. JSON object containing the resolved subject metadata, according to the requested type and expressed in the 
      metadata format defined in [Section 3](https://openid.net/specs/openid-federation-1_0.html#entity-statement).
```

If the requester is to be guaranteed to receive all metadata for the selected type, the federation_entity metadata, if present, must also be returned.

2 possible solutions to this issue is:

- Change the recommendation in 5.2.2 and state that the common attributs must be placed with each entity type they apply to, or
- Enforce that federation_entity metadata is always returned in resolve responses, even if type is provided in the request.

I originally requested the first option (I still like that best), but I have implemented the second to be compliant with the standard and other implementations of it.

selfissued:
Fixes #43 

Cc: @Razumain

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -9476,6 +9476,7 @@ Host: op.umu.se
 	  <t>
 	    Fixed #43: Allow multiple <spanx style="verb">type</spanx> request parameters in resolve requests.
 	  </t>
+    <t>
 	    Fixed #40: Changed section name from "Resolve Entity Statement" to "Resolve Entity".
 	  </t>
 	    Fixed #39: Removed <spanx style="verb">iss</spanx> parameter from fetch endpoint.
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -4091,6 +4091,11 @@ Content-Type: application/json
                 Its value is an Entity Type Identifier, as specified in
                 <xref target="entity_types"/>.
                 If this parameter is not present, then all Entity Types are returned.
+		<vspace blankLine="1"/>
+		The <spanx style="verb">type</spanx> request parameter
+		MAY occur multiple times, in which case, data for each Entity Type
+		whose Entity Type Identifier is a <spanx style="verb">type</spanx>
+		parameter value is returned.
               </t>
             </list>
           </t>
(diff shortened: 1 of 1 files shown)


vdzhuvinov (approved):

Razumain:
The change is fine.
Would you agree to say the following in addition:
"Metadata for federation_entity shall/should/may always be returned regardless of selected types as it may contain metadata that are common to all types." ?

peppelinux (approved):

openid-federation-1_0.xml
```diff
@@ -9469,6 +9474,8 @@ Host: op.umu.se
 	    Endpoint URLs are not form-urlencoded in JSON metadata parameter values.
 	  </t>
 	  <t>
+	    Fixed #43: Allow multiple <spanx style="verb">type</spanx> request parameters in resolve requests.
+	  </t>
 	    Fixed #40: Changed section name from "Resolve Entity Statement" to "Resolve Entity".
```
peppelinux, line 9480:
```suggestion
    <t>
	    Fixed #40: Changed section name from "Resolve Entity Statement" to "Resolve Entity".
```

openid-federation-1_0.xml
```diff
@@ -9469,6 +9474,9 @@ Host: op.umu.se
 	    Endpoint URLs are not form-urlencoded in JSON metadata parameter values.
 	  </t>
 	  <t>
+	    Fixed #43: Allow multiple <spanx style="verb">type</spanx> request parameters in resolve requests.
+	  </t>
+    <t>
 	    Fixed #40: Changed section name from "Resolve Entity Statement" to "Resolve Entity".
 	  </t>
 	    Fixed #39: Removed <spanx style="verb">iss</spanx> parameter from fetch endpoint.
```
peppelinux, line 9483:
```suggestion
     <t>
	    Fixed #39: Removed <spanx style="verb">iss</spanx> parameter from fetch endpoint.
```

rohe (approved):

PR #80, opened 2024-09-07
Explicitly type signed JWK Sets

### LINKED ISSUE #64: signed_jwks_uri and JWS type
In [5.2.1. ](https://openid.net/specs/openid-federation-1_0.html#section-5.2.1)[Extensions for JWK Sets in Entity Metadata](https://openid.net/specs/openid-federation-1_0.html#name-extensions-for-jwk-sets-in-)
 there is a description of 'signed_jwks_uri' which returns a signed JWT with JWKS and other specific claims... 

I would expect that this signed JWT has a [application/jwk-set+jwt](https://openid.net/specs/openid-federation-1_0.html#name-application-jwk-setjwt-medi) type, but I don't see any mention of it. 

Edit: to clarify, the question is - should this Signed JWKS have a typ header of jwk-set+jwt?

selfissued:
Fixes #64 

Cc: @cicnavi

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -1602,12 +1602,23 @@
 		<vspace/>
 		OPTIONAL. URL referencing a signed JWT having the Entity's
 		JWK Set document for that Entity Type as its payload.
-		The JWT MUST be signed using a Federation Entity Key.
 		This URL MUST use the <spanx style="verb">https</spanx> scheme.
-		When both signing and encryption keys are made available,
+		The JWT MUST be signed using a Federation Entity Key.
+		A successful response from the URL MUST use the HTTP status code 200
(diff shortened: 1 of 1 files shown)


vdzhuvinov (approved):

peppelinux (approved):

openid-federation-1_0.xml
```diff
@@ -1706,6 +1717,10 @@
 		OPTIONAL.
 		JSON Web Key Set document, passed by value,
 		containing the Entity's keys for that Entity Type.
+		When both signing and encryption keys are present,
```
peppelinux, line 1720: well, actually in openid federation we don't have encryption ops

rohe (approved):

PR #81, opened 2024-09-07
Define entity_statement_signing_alg_values_supported metadata parameter

### LINKED ISSUE #65: Entity statement signing algorithm
In the spec it is stated about entity statements (in Section 3) :

"Implementations SHOULD support signature verification with the RSA SHA-256 algorithm because OpenID Connect Core requires support for it (alg value of RS256). Federations MAY also specify different mandatory-to-implement algorithms."

This implies that the federation should be able in some way to publish which algorithm that should be used for signing entity statements within the federation.
How this could be done is not discussed in the spec. I think it should be.

I'd also like to propose that the TA should publish which signing algorithm the federation should use together with the its public keys. This would allow the federation to move from one algorithm to another under controlled circumstances.
I make the assumption here that all entities within the federation MUST used the same algorithm.

selfissued:
Fix #65 

Cc: @Razumain

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -466,6 +466,9 @@
           The Entity Statement is signed using one of the private keys of the issuer
           Entity in the form of a <xref target="RFC7515">JSON Web Signature
           (JWS)</xref>.
+	  Signing algorithms supported by the Entity MAY be declared using the
+	  <spanx style="verb">entity_statement_signing_alg_values_supported</spanx>
+	  Federation Entity metadata parameter defined in <xref target="federation_entity"/>.
           Implementations SHOULD support signature verification with
           the RSA SHA-256 algorithm because OpenID Connect Core
           requires support for it (<spanx style="verb">alg</spanx>
@@ -1219,6 +1222,14 @@
           All Federation Entities MAY publish a
 	      <spanx style="verb">federation_historical_keys_endpoint</spanx>.
 	    </t>
+	    <t hangText="entity_statement_signing_alg_values_supported">
+	      <vspace/>
+	      OPTIONAL.
+	      JSON array containing a list of the JWS signing algorithms
+	      (<spanx style="verb">alg</spanx> values)
+	      supported by the Entity to sign Entity Statements.
+	      If omitted, the default value is <spanx style="verb">["RS256"]</spanx>.
+	    </t>
           </list>
         </t>
 	<t>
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -1219,6 +1222,14 @@
           All Federation Entities MAY publish a
 	      <spanx style="verb">federation_historical_keys_endpoint</spanx>.
 	    </t>
+	    <t hangText="entity_statement_signing_alg_values_supported">
+	      <vspace/>
+	      OPTIONAL.
+	      JSON array containing a list of the JWS signing algorithms
+	      (<spanx style="verb">alg</spanx> values)
+	      supported by the Entity to sign Entity Statements.
+	      If omitted, the default value is <spanx style="verb">["RS256"]</spanx>.
```
vdzhuvinov, line 1231: On its own, the definition of this parameter is okay. However, having this parameter parameter defined would not be sufficient for the entities in a federation to be able to use a specific JWS algorithm when the more than one is supported. People will ask - what do we do then with this parameter? How do I get an ES signed with say ES256 when RS256 and ES256 are supported?

vdzhuvinov (changes requested):

peppelinux (approved):

vdzhuvinov:
Last time when we discussed this (and I was present), there were two solutions.

Both solutions have this metadata parameter, that lists the supported JWS algs for entity statements.

The difference is:

1. Entities Statements can be signed with any of the supported JWS algs. There is no method to request the issue of an Entity Statement with a given JWS alg. This means consumers must be prepared to handle all JWS algs that are supported.

2. Entity Statements are normally signed with a default JWS alg, but there is also the possibility to request a specific JWS alg using a query param like `jws_alg=ES256`. With that consumers have a mean to control the signature algorithm.

Solution 1 seems simpler, in terms of specification and API.

Solution 2 has the benefit that in multi anchored / lateral federations, where one of the federations supports additional JWS algs, but the algs still intersect (there is a commonly supported JWS alg, like RS256), consumers will be able to collect Entity Statements with an alg that they can work it. Otherwise they won't be able to do this.

selfissued:
As discussed on today's Federation editors' call, we would want actionable guidance on how and when to use this parameter before including this feature in the specification.

rohe (approved):

selfissued:
Closing on the same basis as #65

PR #91, opened 2024-09-16
Validate explicit registration audiences

### LINKED ISSUE #88: Explicit client registration: Mention that the OP must check the "aud" of the received EC
Mention in section [12.2.2.1. ](https://openid.net/specs/openid-federation-1_0.html#section-12.2.2.1) that OP needs to verify the "aud" claim in the received Entity Statements - that the claim is present and it has a single value that is the OP's Entity Identifier.

This check is required to prevent the misuse of an RP's EC published at its well-known endpoint. In a federation that supports both methods of client registration - automatic and explicit, if the "aud" check isn't made by the OP, one could download the RP's EC and use it to trigger its registration at the OP.

selfissued:
Fixes #88

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -6389,6 +6389,8 @@ HTTP/1.1 302 Found
                   <vspace/>
                   REQUIRED.
                   Its value MUST be the Entity Identifier of the OP.
+		  This claim is used in Explicit Registration requests
+		  but is not a general Entity Statement claim.
                 </t>
                 <t hangText="authority_hints">
                   <vspace/>
@@ -6437,6 +6439,13 @@ HTTP/1.1 302 Found
                     the content type to determine whether it contains an Entity Configuration
                     or an entire Trust Chain.
                   </t>
+		  <t>
+		    The OP MUST validate the RP's explicit registration request JWT.
+		    All the normal Entity Statement validation rules apply.
+		    In addition, if the <spanx style="verb">aud</spanx> (audience)
+		    Claim value is not the Entity Identifier of the OP,
+		    then the request MUST be rejected.
+		  </t>
                   <t>
                     If the request contains an Entity Configuration the OP
                     MUST use it to complete the Federation Entity Discovery by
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -6389,6 +6389,8 @@ HTTP/1.1 302 Found
                   <vspace/>
                   REQUIRED.
                   Its value MUST be the Entity Identifier of the OP.
+		  This claim is used in Explicit Registration requests
+		  but is not a general Entity Statement claim.
```
peppelinux, line 6393:
```suggestion
		  This claim is only used in Explicit Registration requests,
		  since it is not a general Entity Statement claim.
```

openid-federation-1_0.xml
```diff
@@ -6437,6 +6439,13 @@ HTTP/1.1 302 Found
                     the content type to determine whether it contains an Entity Configuration
                     or an entire Trust Chain.
                   </t>
+		  <t>
+		    The OP MUST validate the RP's explicit registration request JWT.
+		    All the normal Entity Statement validation rules apply.
+		    In addition, if the <spanx style="verb">aud</spanx> (audience)
+		    Claim value is not the Entity Identifier of the OP,
```
peppelinux, line 6446:
```suggestion
		    claim value is not the Entity Identifier of the OP,
```

peppelinux (approved):

openid-federation-1_0.xml
```diff
@@ -6610,10 +6619,10 @@ HTTP/1.1 302 Found
...
                     <vspace/>
                     REQUIRED.
-                    Its value MUST be the Entity Identifier of the RP. See
-                    <xref target="RFC7519"/> for the full specification. This
-                    claim is specific to Explicit Registration responses and is
-                    not a general Entity Statement claim.
+                    Its value MUST be the Entity Identifier of the RP.
+		    See <xref target="RFC7519"/> for the full specification.
+		    This claim is used in Explicit Registration responses
+		    but is not a general Entity Statement claim.
```
vdzhuvinov, line 6625: To keep in mind that this issue relates to the "aud" in Entity Statement claims in response to authenticated callers (with `private_key_jwt`):

https://github.com/openid/federation/issues/98

vdzhuvinov (approved):

PR #95, opened 2024-09-21
Correct aud example

### LINKED ISSUE #90: Inconsistent example regarding Request Object in automatic registration
In [12.1.1.1. ](https://openid.net/specs/openid-federation-1_0.html#section-12.1.1.1)[Using a Request Object](https://openid.net/specs/openid-federation-1_0.html#name-using-a-request-object) description of `aud` claim says that it must be OP's Issuer Identifier URL. 

> aud
REQUIRED. The Audience (aud) value MUST be or include the OP's Issuer Identifier URL

In example in [12.1.1.1.1. ](https://openid.net/specs/openid-federation-1_0.html#section-12.1.1.1.1)[Authorization Request with a Trust Chain](https://openid.net/specs/openid-federation-1_0.html#name-authorization-request-with-) for `aud` claim is given:

> "aud": "https://op.example.org/authorization-endpoint"

which seems as if authorization endpoint is used.

selfissued:
Fixes #90

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -5919,7 +5919,7 @@ Content-Type: application/json
 }
 .
 {
-  "aud": "https://op.example.org/authorization-endpoint",
+  "aud": "https://op.example.org",
   "client_id": "https://rp.example.com",
   "exp": 1589699162,
   "iat": 1589699102,
@@ -9825,7 +9825,7 @@ Host: op.umu.se
 	-40
 	<list style="symbols">
 	  <t>
-	    TBD
+	    Fixed #90: Corrected request object audience example.
 	  </t>
 	</list>
       </t>


cicnavi (approved):

openid-federation-1_0.xml
```diff
@@ -9825,7 +9825,7 @@ Host: op.umu.se
 	-40
 	<list style="symbols">
 	  <t>
-	    TBD
+	    Fixed #90: Corrected request object audience example.
```
peppelinux, line 9828: since it was for a non normative example, I would keep it out from the changelog

peppelinux (approved):

openid-federation-1_0.xml
selfissued, line 9828:
```suggestion
	    TBD
```

rohe (approved):

vdzhuvinov (approved):

PR #96, opened 2024-09-22
Describe validation of resolved metadata

### LINKED ISSUE #28: Federation: Validation of metadata in entity statement
This question should be read in the light of conformance testing of federation entities:

Let's say that we've got an entity statement that contains the following metadata:

```
{
  "federation_entity": {
    "federation_fetch_endpoint": "https://example.com/fetch",
    "federation_list_endpoint": "https://example.com/list"
  },
  "oauth_authorization_server": {},
  "openid_relying_party": {}
}
```

`federation_entity` is valid and included for illustrative purposes. But what about `oauth_authorization_server` and `openid_relying_party`?

On one hand, the spec says that

> When an Entity participates in a federation or federations with one or more Entity Types, its Entity Configuration MUST contain a metadata claim with JSON object values for each of the corresponding Entity Type Identifiers, even if the values are the empty JSON object {} 

but on the other hand, if we take openid_relying_party as an example, it says

> All parameters defined in Section 2 of [OpenID Connect Dynamic Client Registration 1.0](https://openid.net/specs/openid-federation-1_0.html#OpenID.Registration) [[OpenID.Registration](https://openid.net/specs/openid-federation-1_0.html#OpenID.Registration)] and [Section 5.2](https://openid.net/specs/openid-federation-1_0.html#common_metadata) are applicable, as well as additional parameters registered in the IANA "OAuth Dynamic Client Registration Metadata" registry [[IANA.OAuth.Parameters](https://openid.net/specs/openid-federation-1_0.html#IANA.OAuth.Parameters)]. 

But what does “applicable” mean? Does it mean that all of the properties and their optionality (or lack thereof) as defined in the DCR spec apply, meaning that we should run the same set of verifications that we do for RP DCR client registration metadata, or does it mean that the properties in the DCR spec may occur inside this metadata section, but are not subject to any validation rules?

If you could please enlighten me a little bit in this matter, that would be helpful.

**Edit:** I left out the little detail that client_registration_types is required in the openid_relying_party metadata section, but I guess that isn’t really relevant to the question anyway and that the description above explains it sufficiently as it is.

selfissued:
Fixes #28

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -1098,7 +1098,7 @@
     <section anchor="metadata" title="Metadata">
       <t>
 	This section defines how to represent and use metadata about Entities.
-	It reuses existing OpenID Connect and OAuth 2.0 metadata standards
+	It uses existing OpenID Connect and OAuth 2.0 metadata standards
 	that are applicable to each Entity Type.
       </t>
       <t>
@@ -1339,7 +1339,7 @@
 	</figure>
       </section>
 
-      <section title="OpenID Provider" anchor="OP_metadata">
+      <section title="OpenID Connect OpenID Provider" anchor="OP_metadata">
         <t>
           The Entity Type Identifier is
           <spanx style="verb">openid_provider</spanx>.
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -2847,16 +2847,20 @@
...
 
               <t>
                 The Trust Chain subject is responsible to verify that it is
-                able to support and comply with the metadata that results from
+                able to support and comply with the resolved metadata that results from
                 the application of federation metadata policies. For instance,
                 this may involve a check that cryptographic algorithms required
-                by the resulting metadata are supported. When metadata policies
+                by the resulting resolved metadata are supported.
+		Likewise, the Entity MUST verify that
```
vdzhuvinov, line 2860: the Entity -> the Trust Chain subject

openid-federation-1_0.xml
```diff
@@ -2847,16 +2847,20 @@
...
+		all the required metadata parameters for its Entity Type are present
```
vdzhuvinov, line 2861: Entity Type -> Entity Types

vdzhuvinov (changes requested):
Just minor comment to clarify who is doing the resolved metadata validation :)

peppelinux (approved):

MichaelFraser1999 (approved):

Issue #94, opened 2024-09-20
Federation Analysis: Private Key JWT Replay


selfissued:
From: Pedram Hosseyni <[pedram.hosseyni@sec.uni-stuttgart.de](mailto:pedram.hosseyni@sec.uni-stuttgart.de)> 
Sent: Friday, September 20, 2024 2:48 AM
To: [michael_b_jones@hotmail.com](mailto:michael_b_jones@hotmail.com)
Cc: Marcus Almgren <[marcus.almgren@oidf.org](mailto:marcus.almgren@oidf.org)>; Ralf Kuesters <[ralf.kuesters@sec.uni-stuttgart.de](mailto:ralf.kuesters@sec.uni-stuttgart.de)>; Tim Würtele <[tim.wuertele@sec.uni-stuttgart.de](mailto:tim.wuertele@sec.uni-stuttgart.de)>
Subject: Federation Analysis: Private Key JWT Replay

Hi Mike,

For the Federation analysis, we found a scenario that seems to be in line with the specifications and would lead to attacks.

To put it briefly, Federation uses PAR for automatic registration, so we consider OIDC together with PAR, on top of Federation for registration. 
However, the PAR specification allows for (1) redirection endpoints that are not pre-registered and (2) using the token endpoint as the audience value of the private key JWT when authenticating at the PAR endpoint. 
However, a malicious OP could specify token endpoints of other OPs, thus, obtaining private key JWTs created by an RP that it could use at those OPs.

As the Federation specification uses PAR, we think that it should also provide enough restrictions to ensure that the resulting protocol is secure; For example, Federation already introduces restrictions on the set of authentication mechanisms that should be used for PAR.

The attached PDF contains a detailed description, including diagrams and our assumptions.

Best
Tim and Pedram

Cc: @PedramHD @SECtim

selfissued:
[2024-09-20-private-key-JWT-attack.pdf](https://github.com/user-attachments/files/17082698/2024-09-20-private-key-JWT-attack.pdf)

Razumain:
Just saying.

This is yet another reason why I think OpenID federation should be limited or split. I think 2 documents would be fine.

OpenID federation should only focus on the federation aspect. That is to share validated metadata and trust marks through the current structure. And that's it.

This can then be used alone or together with profiles that defines rules for registration and authentication requests.

[...]

PR #99, opened 2024-09-29
Simplified use of PAR for request authentication

### LINKED ISSUE #34: Drawbacks of defining request_authentication_methods_supported as JSON object
When defining `request_authentication_methods_supported` as JSON object, the document changes a long tradition of not doing so, which I think has significant drawbacks.

Traditionally, metadata profiles makes great effort to define metadata as string, integer, boolean or an array of the former in some form.

This creates may metadata parameters, but it also has the advantage of making it easy to process metadata parameters as well as allowing effective policy processing.

We can now see that objects are added as optional type of metadata values, but I have serious doubts that this will work in practice. To really implement policy processing of metadata values in JSON object form will be a huge challenge.

Taking token_endpoint as example.
The following has already been defined elsewhere:

>`token_endpoint`
>`token_endpoint_auth_method`

Wouldn't it then be better and more consistent to define:

>`token_endpoint_request_auth_methods_supported`

And then do the same for other endpoint types.
Here are the advantages I can see:

1. Today the name of the endpoint is an arbitrary string. It requires definition of the names of each endpoint. By this change, that is not required.
2. Staying in line with how metadata properties are defined in general.
3. These metadata values can be influenced by policy in a much more straightforward fashion.
4. Standard tools for processing metadata values can be used without modification. Easier development and better backwards compatibility with existing tools.

### LINKED ISSUE #98: Revisit private_key_jwt client auth at the federation endpoints ("aud" claim)
Several months back we specced the possibility for callers to authenticate at federation endpoints with `private_key_jwt`.

**8.8 Client Authentication at Federation Endpoints**
https://openid.net/specs/openid-federation-1_0.html#section-8.8

In view of the recent sec review, two points to consider:

1. Change the `aud` from the endpoint URL to the Entity ID?
2. Should JWT responses to authenticated requests set the `aud`? Reasons why one may want to "aud" the response? Pros / cons? Requiring the `aud` to be set will largely prevent the issuing Entity from being able to cache the minted JWTs.

selfissued:
Fixes #34 
Replaces `request_authentication_methods_supported` with `pushed_authentication_methods_supported`.  This simplifies things by eliminating the `request_object` request authentication method, since signed requests are always required.  The PR also removes `tls_client_auth` from the supported request authentication methods.

Fixes #98 
Requires the audience when using `private_key_jwt` to be the Authorization Server's Entity Identifier.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -1407,21 +1407,6 @@
               Registration as described in <xref target="explicit"/>,
               then this claim is REQUIRED.
             </t>
-
-            <t hangText="request_authentication_signing_alg_values_supported">
-              <vspace/>
-              OPTIONAL.
-              JSON array containing a list of the supported
-              <xref target="RFC7515">JWS</xref> algorithms
-              (<spanx style="verb">alg</spanx> values)
-              for signing the <xref target="RFC7519">JWT</xref>
-              used in the Request Object of the OpenID Connect authentication request
-	      or in the <spanx style="verb">private_key_jwt</spanx> JWT
-	      of a pushed authorization request to the OP.
-              No default algorithms are implied if this entry is omitted.
-              Servers SHOULD support <spanx style="verb">RS256</spanx>.
-              The value <spanx style="verb">none</spanx> MUST NOT be used.
-            </t>
           </list>
         </t>
 
@@ -1476,7 +1461,11 @@
             "private_key_jwt"
          ],
          "pushed_authorization_request_endpoint":"https://op.umu.se/openid/par",
-         "request_authentication_signing_alg_values_supported": [
+         "request_object_signing_alg_values_supported": [
+            "ES256",
+            "RS256"
+         ],
+         "token_endpoint_auth_signing_alg_values_supported": [
             "ES256",
             "RS256"
          ]
@@ -7139,25 +7128,6 @@ HTTP/1.1 302 Found
           <t>
             <list style='symbols'>
               <t>
-                Metadata Name: <spanx style="verb">request_authentication_signing_alg_values_supported</spanx>
-              </t>
-              <t>
-                Metadata Description:
-                JSON array containing the JWS signing algorithms
-                supported for the signature on
-                the JWT used to authenticate the request
-              </t>
-              <t>
-                Change Controller: OpenID Foundation Artifact Binding Working Group - openid-specs-ab@lists.openid.net
-              </t>
-              <t>
-                Specification Document(s): <xref target="OP_metadata"/> of this specification
-              </t>
-            </list>
-          </t>
-          <t>
-            <list style='symbols'>
-              <t>
                 Metadata Name: <spanx style="verb">signed_jwks_uri</spanx>
               </t>
               <t>
@@ -9742,8 +9712,10 @@ Host: op.umu.se
 	  </t>
 	  <t>
 	    Fixed #34: Deleted
-	    <spanx style="verb">request_authentication_methods_supported</spanx>
-	    and replaced with the use of standard PAR metadata values.
+	    <spanx style="verb">request_authentication_methods_supported</spanx> and
+	    <spanx style="verb">request_authentication_signing_alg_values_supported</spanx>
+	    and replaced with the use of standard Request Object
+	    and PAR metadata values.
 	    Also restricted PAR authentication methods to those
 	    performing signing with the RP's keys.
 	  </t>
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -5987,7 +5987,7 @@ GET /authorize?
             <t>
               <xref target="RFC9126">Pushed Authorization Requests</xref>
               provide an
-              interoperable way to push an authentication request parameters
+              interoperable way to push authentication request parameters
               directly to the AS in exchange for a one-time-use
               <spanx style="verb">request_uri</spanx>.
 	      The standard PAR metadata parameters are used in the
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -1232,6 +1232,21 @@
           All Federation Entities MAY publish a
 	      <spanx style="verb">federation_historical_keys_endpoint</spanx>.
 	    </t>
+
+            <t hangText="endpoint_authentication_signing_alg_values_supported">
+              <vspace/>
+              OPTIONAL.
+              JSON array containing a list of the supported
+              <xref target="RFC7515">JWS</xref> algorithms
+              (<spanx style="verb">alg</spanx> values)
+              for signing the <xref target="RFC7519">JWT</xref>
+              used for <spanx style="verb">private_key_jwt</spanx>
+	      when authenticating to federation endpoints,
+	      as described in <xref target="ClientAuthentication"/>.
+              No default algorithms are implied if this entry is omitted.
+              Servers SHOULD support <spanx style="verb">RS256</spanx>.
+              The value <spanx style="verb">none</spanx> MUST NOT be used.
+            </t>
           </list>
         </t>
 	<t>
@@ -1379,44 +1394,6 @@
               then this claim is REQUIRED.
             </t>
 
-            <t hangText="pushed_authentication_methods_supported">
-              <vspace/>
-              OPTIONAL.
-	      A Pushed Authorization Request (PAR) <xref target="RFC9126"/>
-	      can be used with Automatic Registration,
-	      as described in <xref target="authn-request"/>.
-	      Some OAuth 2.0 client authentication methods can be used at the
-	      <spanx style="verb">pushed_authorization_request_endpoint</spanx>
-	      to achieve request authentication.
-	      They are the subset of those listed in the
-	      IANA "OAuth Token Endpoint Authentication Methods" registry <xref target="IANA.OAuth.Parameters"/>
-	      that perform client authentication by proving possession of a private key.
-	      The <spanx style="verb">pushed_authentication_methods_supported</spanx>
-	      metadata value lists the client authentication methods
-	      supported by the OP that can be used in this manner.
-	      Its value is a JSON array of supported client authentication methods.
-	      Values that MAY be used are:
-              <list style="hanging">
-                <t hangText="private_key_jwt">
-                  <vspace/>
-                  This client authentication method is described in Section 9 of
-                  <xref target="OpenID.Core">OpenID Connect Core 1.0</xref>.
-                  When <spanx style="verb">private_key_jwt</spanx> is used,
-		  the audience of the signed JWT MUST be
-                  the Authorization Server's Entity Identifier.
-                </t>
-                <t hangText="self_signed_tls_client_auth">
-                  <vspace/>
-                  This client authentication method is described in Section 2.2 of
-                  <xref target="RFC8705"/>.
-                </t>
-              </list>
-	      <vspace blankLine="1"/>
-	      In both cases, the signing key MUST be one of the keys
-	      in the RP's JWK Set published in its metadata.
-	      Additional pushed authentication methods MAY be defined and used.
-	    </t>
-
             <t hangText="request_authentication_signing_alg_values_supported">
               <vspace/>
               OPTIONAL.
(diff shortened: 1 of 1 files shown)


peppelinux (approved):

openid-federation-1_0.xml
```diff
@@ -1378,77 +1378,43 @@
...
-                  <xref target="automatic"/>.
-                </t>
-              </list>
-	      The <spanx style="verb">request_object</spanx> method MAY be used at either endpoint.
-              <vspace blankLine="1"/>
-
-	      Additional request authentication methods MAY be defined and used.
+	      <vspace blankLine="1"/>
+	      In both cases, the signing key MUST be one of the keys
+	      in the RP's JWK Set published in its metadata.
```
SECtim, line 1416: It may be helpful to clarify that the keys referenced here are OIDC RP keys (i.e., published under `metadata`->`openid_relying_party`, and NOT Federation keys).

openid-federation-1_0.xml
```diff
@@ -1378,77 +1378,43 @@
...
+	      Additional pushed authentication methods MAY be defined and used.
```
SECtim, line 1417: Any "additional pushed authentication methods" used with PAR in automatic registration should (or must?) rely on some kind of asymmetric key pair where the OP learns the RP's public key from the RP's metadata (as resolved via Federation). However, such restrictions of Federation extension points might be considered to be out of scope, I just wanted to bring it up for discussion.

openid-federation-1_0.xml
```diff
@@ -1378,77 +1378,43 @@
...
-                  the Authorization Server's Authorization Endpoint or the
-                  Authorization Server's Entity Identifier.
-                </t>
-                <t hangText="tls_client_auth">
-                  <vspace/>
-                  Section 2.1 of
-                  <xref target="RFC8705"/>.
+                  When <spanx style="verb">private_key_jwt</spanx> is used,
+		  the audience of the signed JWT MUST be
+                  the Authorization Server's Entity Identifier.
```
SECtim, line 1406: Does this restriction also apply to Federation deployments that do not use the `pushed_authentication_methods_supported` metadata parameter, but nonetheless still support PAR? (since the metadata parameter is marked as optional)

openid-federation-1_0.xml
```diff
@@ -1378,77 +1378,43 @@
               Registration as described in <xref target="explicit"/>,
               then this claim is REQUIRED.
             </t>
-            <t hangText="request_authentication_methods_supported">
+
+            <t hangText="pushed_authentication_methods_supported">
               <vspace/>
               OPTIONAL.
```
SECtim, line 1384: The whole `pushed_authentication_methods_supported` metadata parameter is currently marked as optional. Would it make sense to recommend (or even require) this metadata parameter for all OIDC OPs that support PAR (or at least: that support PAR for registration as described in Section 12, similar to the `federation_registration_endpoint`)?

openid-federation-1_0.xml
```diff
@@ -9825,7 +9777,13 @@ Host: op.umu.se
...
 	<list style="symbols">
 	  <t>
-	    TBD
+	    Fixed #34:
+	    Replaced <spanx style="verb">request_authentication_methods_supported</spanx>
+	    with <spanx style="verb">pushed_authentication_methods_supported</spanx>.
+	  </t>
+	  <t>
+	    Fixed #98: Required audience when using <spanx style="verb">private_key_jwt</spanx>
+	    to be the Authorization Server's Entity Identifier.
```
SECtim, line 9786: Currently, this requirement only applies to pushed authorization requests, not to token requests, and not to requests to Federation endpoints with client authentication.

SECtim (changes requested):
Thanks @selfissued for drafting this PR!

- While I still have to discuss this with @PedramHD, but I think this PR might also prevent the attack described in #94 (if (1) the pkjwt restriction introduced in this PR applies to all Federation-using OPs, see also my comments on the proposed changes. And (2) at least in the scenario considered in #94, i.e., Federation + OIDC without further extensions; in particular, no endpoints at which an RP needs to authenticate except for PAR, Token, and optionally Federation endpoints). However, I still think #94 merits at least some additional security considerations.
- (requested change) I think this PR should also change the following list for consistency
https://github.com/openid/federation/blob/d0166f6cb5fe3d89b5906fd1b10f5163901989a2/openid-federation-1_0.xml#L6052-L6082
- In its current state, I am unsure how this addresses #98, since this PR does not seem to change anything related to authentication at Federation endpoints.

vdzhuvinov:
I'm wondering, can we simplify things further? 

What if there were no `*_authentication_methods_supported` parameters at all?

Consider the following OP metadata, which is entirely based on OIDC Discovery:
```
{
   "request_parameter_supported": true,
   "request_object_signing_alg_values_supported": ["RS256", "ES256"]
}
```
A Federation RP will deduce from this that it can make an OpenID authentication request with an RS256 signed object and that because the OP is Federation enabled, this request should work for automatic Federation RPs also (provided the request object has the extra Federation specific JWT claims of course).

Or the OP metadata:
```
{
   "pushed_authorization_request_endpoint": "https://op.example.com/par",
   "require_pushed_authorization_requests": true,
   "token_endpoint_auth_methods_supported": "private_key_jwt",
   "token_endpoint_auth_signing_alg_values_supported": ["RS256", "ES256"]
}
```

A Federation RP will deduce that:

- The OP supports and requires PAR (i.e. regular, including signed authZ requests, will be rejected - this could be a FAPI 2.0 deployment for instance)
- The RP must authenticate with RS256 private_key_jwt
- Because the above capability is advertised, it should also work for Federation automatic clients.

Here are the only reasons why an OP may want to have a separate metadata parameter, i.e. the `*_authentication_methods_supported`:

- The OP supports both automatic and explicit client registration, and for some reason it doesn't implement or want to allow auto registration for one / some of the authZ request variants.
- Like the previous point, but the OP is also having non-Federation clients.

vdzhuvinov:
> This simplifies things by eliminating the request_object request authentication method, since signed requests are always required

OPs that implement PAR and have `"require_pushed_authorization_requests": true` will accept neither plain nor signed authorisation requests. FAPI 2.0 for instance mandates this. I realised that because of this, the `request_object` cannot be an MTI for OPs.

selfissued:
I simplified things further, as suggested by @vdzhuvinov.  The federation-specific PAR metadata is now gone.

openid-federation-1_0.xml
```diff
@@ -6057,17 +5982,28 @@ GET /authorize?
...
             </t>
+	    <t>
+	      When used with OpenID Federation,
+	      a Request Object MUST be sent to the PAR endpoint,
+	      signed with one of the RPs private keys.
+	      The corresponding public key MUST be in the RP's JWK Set.
+	      Furthermore, when used with OpenID Federation,
+	      the client authentication methods for the PAR endpoint
+	      are restricted to those proving possession of a private key.
+	      Likewise, the corresponding public key MUST be in the RP's JWK Set.
```
SECtim, line 5999: It may be helpful to clarify that the keys referenced here are OIDC RP keys (i.e., published under metadata>openid_relying_party, and NOT Federation keys).

SECtim (approved):
Most of my previous comments are now obsolete due to the removal of the PAR-related metadata or addressed.
Also, #103 fixes #98 (which this PR does not), so that comment is also obsolete.

vdzhuvinov (changes requested):

openid-federation-1_0.xml
```diff
@@ -1386,78 +1401,6 @@
...
-                  <xref target="automatic"/>.
-                </t>
-              </list>
-	      The <spanx style="verb">request_object</spanx> method MAY be used at either endpoint.
-              <vspace blankLine="1"/>
-
-	      Additional request authentication methods MAY be defined and used.
-	    </t>
 
             <t hangText="request_authentication_signing_alg_values_supported">
```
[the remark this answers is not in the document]
vdzhuvinov: For request objects the RP can find out the supported JWS algs from the OP `request_object_signing_alg_values_supported` metadata.

When using `private_key_jwt` at the PAR endpoint the RP can find out the supported JWS algs from the OP `token_endpoint_auth_signing_alg_values_supported` metadata. For client auth at the PAR endpoint we literary take what's specified for the token endpoint:

https://datatracker.ietf.org/doc/html/rfc9126#name-request

> The authorization server MUST process the request as follows:
> 1. Authenticate the client in the same way as at the token endpoint ([Section 2.3](https://www.rfc-editor.org/rfc/rfc6749#section-2.3) of [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)]).

openid-federation-1_0.xml
```diff
@@ -6057,17 +5982,28 @@ GET /authorize?
...
-              directly to the AS in exchange for a
+              interoperable way to push an authorization request
+              directly to the AS in exchange for a one-time-use
               <spanx style="verb">request_uri</spanx>.
+	      The standard PAR metadata parameters are used in the
+	      RP and OP metadata to indicate its use.
             </t>
+	    <t>
+	      When used with OpenID Federation,
+	      a Request Object MUST be sent to the PAR endpoint,
```
vdzhuvinov, line 5993: Request objects were not required for the PAR endpoint in Federation. Federation clients authenticate at the PAR endpoint as they would at the token endpoint, with `private_key_jwt` or `self_signed_tls_client_auth`.

In the PAR RFC request objects are optional and we should keep that in Federation.

> The endpoint MAY also support sending the set of authorization request parameters as a Request Object according to [[RFC9101](https://datatracker.ietf.org/doc/html/rfc9101)] and [Section 3](https://datatracker.ietf.org/doc/html/rfc9126#request_parameter) of this document.

rohe (approved):
I approve provided that @vdzhuvinov comments are accepted.

openid-federation-1_0.xml
```diff
@@ -6058,85 +5987,71 @@ GET /authorize?
             <t>
               <xref target="RFC9126">Pushed Authorization Requests</xref>
               provide an
-              interoperable way to push a Request Object
-              directly to the AS in exchange for a
+              interoperable way to push an authentication request parameters
```
SECtim, line 5990: Typo in
> interoperable way to push `an` authentication request parameters

(highlight by me)

openid-federation-1_0.xml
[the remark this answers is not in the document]
selfissued: [Deleted request_authentication_signing_alg_values_supported](https://github.com/openid/federation/pull/99/commits/35135f983ac1bd4d2b9dcf3936994d175cbd5fc3)

openid-federation-1_0.xml
selfissued: `pushed_authentication_methods_supported` has been removed.

openid-federation-1_0.xml
selfissued: I believe I'd updated the audience language and examples everywhere.

openid-federation-1_0.xml
```diff
@@ -5911,7 +5828,7 @@ Content-Type: application/json
               <t>
               Due to the large size of a Trust Chain, it may be necessary
               to use the HTTP POST method,
-              <spanx style="verb">request_uri</spanx>,
+              a <spanx style="verb">request_uri</spanx>,
```
vdzhuvinov, line 5831: The `request_uri` asks the OP to fetch an external resource and this makes it easier to launch DoS attacks against Federation OPs. I propose to say `redirect_uri` MUST NOT or SHOULD NOT be used by RPs and accepted by OPs.

OIDC recognised this issue and because of that it has the OP `request_uri_parameter_supported` + `require_request_uri_registration` metadata and the RP `request_uris` metadata. So that the request_uri's can be screened before the RP can use them.

In automatic registration if a `request_uri` is used the OP must trust it blindly. If the downloaded JWT contains no Trust Chain what should the OP do then? Stop? Or resolve the `client_id`, fetching more resources?

It's good security policy to limit and minimise the number of resources at unverified URLs that need to be fetched. POSTing a `request` or using PAR should be sufficient.

vdzhuvinov (changes requested):
Let's consider disallowing `request_uri` when request objects are used. For the why - see my comment below.

PedramHD (approved):

Issue #24, opened 2024-07-29
The iat parameter in Trust Mark status requests creates great overhead for no good reason


Razumain:
The Trust Mark status endpoint takes the issue time as an input parameter.

As the text is written now, the Trust Mark issuer would need to keep track of all issued Trust Marks.

In our implementation, we will issue a new Trust Mark whenever asked, without regard to whether a previous one already exists. There is a very good reason for that. This avoids a complicated algorithm that decides how long a valid TrustMark must be valid in the future in order to deliver it. If we don’t we could ship an existing TrustMark that will expire 1s later, and that is not very useful. By issuing a new one when requested, we make sure that it has the standard lifespan when it is sent.

In order to correctly process the iat parameter, we would have to store issue times for every TrustMark ever issued. And the response will not be very useful if the subject has the requested Trust Mark, but that particular JWT has expired.

In our usage, all the requester is interested in, is if the subject has this TrustMark at this moment in time. No more and no less. We do not plan to store all issued versions of that Trust Mark. It is the verifiers responsibility to check the expiration time. They don’t need the validation service to do that.

I see no legitimate reason for the iat parameter and suggest that it is removed. The status indication is not even assuming the latest issued Trust Mark. This would give a strange result if a status request would be sent for a subject who has that Trust Mark, but where no one has requested a JWT for it in a while. The status response would say false (The latest JWT has expired), but the subject has the Trust Mark, and a new fresh one can be requested.

At least I would request that The Trust Mark issuer is allowed to ignore this parameter if received.

Currently we have to reject all requests that includes iat.

Razumain:
That is part of the problem. There is no discussion about this, but a reader can hardly draw any other conclusion.

If you don't keep track of previously issued Trust Mark, then what are you supposed to do if the "iss" parameter is provided?

The text also says that if "iss" is not provided, then the latest Trust Mark is assumed. That surely also indicate that you need to keep track of issued Trust Marks to know which is last.

I lack a clear description of what a user of this endpoint can expect from a response. Is this a request for the status of a subject in relation to a Trust Mark (i.e. does this subject have this Trust Mark) disregarding any particular issued Trust Mark JWT, or is this a validation service for a particular Trust Mark JWT (Like a X.509 certificate revocation service)?

If this the former (does the subject have this Trust Mark), then "iss" (or "trust_mark") makes no sense. The requester can easily check if the JWT they have has expired, they don't need the validation service to tell them that.

[...]

PR #109, opened 2024-10-09
remove iat and trust_mark parameters from trust mark status endpoint; #24

zachmann:
To move the discussion on #24 forward I submit a proposal to remove the `iat` and `trust_mark` parameters from the trust mark status endpoint.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -4489,12 +4489,10 @@ Host: openid.sunet.se
 		Trust Mark Status Request
 	      </name>
               <artwork><![CDATA[
-POST /federation_trust_mark_status_endpoint HTTP/1.1
+GET /federation_trust_mark_status_endpoint?
+sub=https%3A%2F%2Fopenid.sunet.se%2FRP&
+trust_mark_id=https%3A%2F%2Frefeds.org%2Fsirtfi HTTP/1.1
 Host: op.example.org
-Content-Type: application/x-www-form-urlencoded
-
-sub=https%3A%2F%2Fopenid.sunet.se%2FRP
-&trust_mark_id=https%3A%2F%2Frefeds.org%2Fsirtfi
 ]]></artwork>
             </figure>
         </section>
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -4445,7 +4445,7 @@ Host: openid.sunet.se
       <section title="Trust Mark Status" anchor="status_endpoint">
         <t>
           This enables an Entity to check whether a Trust Mark has been issued to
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -4444,8 +4444,8 @@ Host: openid.sunet.se
 
       <section title="Trust Mark Status" anchor="status_endpoint">
         <t>
-          This enables an Entity to check whether a Trust Mark is still
-          active or not. The query MUST be sent to the Trust Mark issuer.
+          This enables an Entity to check whether a Trust Mark has been issued to
+          an Enity and is still active. The query MUST be sent to the Trust Mark issuer.
```
peppelinux, line 4448:
```suggestion
          an Enitty and is still active. The query MUST be sent to the Trust Mark issuer.
```

zachmann: Almost correct:
```suggestion
          an Entity and is still active. The query MUST be sent to the Trust Mark issuer.
```

selfissued:
The PR changes the HTTP method from POST to GET, but doesn't update the accompanying example, which current is:
> POST /federation_trust_mark_status_endpoint HTTP/1.1
> Host: op.example.org
> Content-Type: application/x-www-form-urlencoded

> sub=https%3A%2F%2Fopenid.sunet.se%2FRP
> &trust_mark_id=https%3A%2F%2Frefeds.org%2Fsirtfi

Please do so.

openid-federation-1_0.xml
```diff
@@ -4458,7 +4458,7 @@ Host: openid.sunet.se
 
         <section anchor="tm-status-request" title="Trust Mark Status Request">
           <t>
-            The request MUST be an HTTP request using the POST method
+            The request MUST be an HTTP request using the GET method
```
peppelinux, line 4461: I ask to support both GET and POST to reduce breaking changes in the current implementations.

technically, from an implementation perspective, this is not a big change.

peppelinux (changes requested):

openid-federation-1_0.xml
selfissued: @zachmann can you please apply @peppelinux's suggestion about supporting both GET and POST?

vdzhuvinov: If Trust Marks a re no longer being submitted, and we're left only with `sub` and `trust_mark_id`, I don't see reason to support POST in the long term.

peppelinux (approved):

openid-federation-1_0.xml
peppelinux: ok

vdzhuvinov (approved):

selfissued (approved):

PR #110, opened 2024-10-09
clarify merging of essential policy operators

zachmann:
As noted in my issue #11 I think the current phrasing on merging the `essential` policy operator is not clear enough.

This is my proposal on how to simplify the phrasing, as I think the semantic should be.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -2494,17 +2494,9 @@
                 Order of application: Last
               </t>
               <t>
-                Operator value merge: If a Superior has specified
-                <spanx style="verb">essential=true</spanx>, then a Subordinate
-                MUST NOT change that. If a Superior has specified
-                <spanx style="verb">essential=false</spanx>, then a Subordinate
-                is allowed to change that to
-                <spanx style="verb">essential=true</spanx>. If a Superior has
-                not specified <spanx style="verb">essential</spanx>, then a
-                Subordinate can set <spanx style="verb">essential</spanx> to
-                <spanx style="verb">true</spanx> or
-                <spanx style="verb">false</spanx>. If those conditions are not
-                met, this MUST result in a policy error.
+                Operator value merge: The result of merging the values of two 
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -2494,17 +2494,9 @@
...
-                is allowed to change that to
-                <spanx style="verb">essential=true</spanx>. If a Superior has
-                not specified <spanx style="verb">essential</spanx>, then a
-                Subordinate can set <spanx style="verb">essential</spanx> to
-                <spanx style="verb">true</spanx> or
-                <spanx style="verb">false</spanx>. If those conditions are not
-                met, this MUST result in a policy error.
+                Operator value merge: The result of merging the values of two 
+                <spanx style="verb">essential</spanx> operators is the logical 
+                disjunction (<spanx style="verb">OR</spanx>) of the operator values.
```
vdzhuvinov, line 2499: The proposed replacement is not equivalent. 

The (inclusive) logical disjunction of:

- Superior policy: essential=true
- Subordinate policy: essential=false

is: essential=true

For the above condition the spec (last sentence in paragraph) describes a policy violation error, not a successful merge.

zachmann: I wasn't completely sure what should be done in this case, but I agree that as it is currently phrased this is the most likely meaning (and you confirmed it).

However, the spec also states that
> If the essential operator is omitted, this is equivalent to including it with a value of `false`. 

Therefore, I believe my suggestion is the only feasible approach.  Otherwise either omitting is not the same as providing `essential=false`, or `essential=true` could not be merged, only if all entities provide `essential=true`.

vdzhuvinov: This shows that we have a bug in the `essential` spec. Good catch!

Would you be able to add the customary history entry to this PR?
Examples here:
https://openid.net/specs/openid-federation-1_0-39.html#appendix-D

selfissued (approved):

peppelinux (approved):

PR #111, opened 2024-10-09
Policy operators: more permissive restrictions when combining add and superset_of

zachmann:
As noted in #11 I'd argue that the `add` and `superset_of` policy operators can be combined nevertheless of their value.
`add` is applied first, then `superset_of` checks if the value is a superset.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -2177,10 +2177,7 @@
                     <spanx style="verb">subset_of</spanx>.
                   </t>
                   <t>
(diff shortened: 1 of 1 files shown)


selfissued:
This was discussed on the 9-Oct-24 Federation editors' call.  @vdzhuvinov intends to look into how this interacts with the defined order of applying policy operators.

selfissued (approved):

selfissued:
Closing since this is replaced by #177

PR #118, opened 2024-10-26
Restrict aud values to Entity Identifier of recipient

### LINKED ISSUE #115: Clarify audience requirements for senders and receivers
Clarify that the audience sent must be a single value containing Entity Identifier of the recipient and that the only acceptable audience value to a recipient is its Entity Identifier, with no other audience values.

@jogu let us know that clarifying this will aid in effective certification testing.

selfissued:
Fixes #115

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -24,7 +24,7 @@
 
   <front>
     <title abbrev="OpenID Federation">OpenID Federation 1.0 -
-      draft 40
+      draft 41
     </title>
 
     <author fullname="Roland Hedberg" initials="R." role="editor"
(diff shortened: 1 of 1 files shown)


rohe (approved):

SECtim (approved):
I did not (yet) thoroughly check that no place was missed, but the proposed changes as such look good to me :+1:

vdzhuvinov:
> I did not (yet) thoroughly check that no place was missed, but the proposed changes as such look good to me 👍

I double checked all "aud" occurrences and they are all covered:

* `private_key_jwt` auth at federation endpoints
* resolve response JWT
* auto registration with PAR + private_key_jwt
* auto registration with `request` object
* explicit registration response

PedramHD (approved):

PR #119, opened 2024-10-26
Describe why and how to not support request_uri

### LINKED ISSUE #114: Disallow request_uri in OpenID authentication requests with automatic registration
> Due to the large size of a Trust Chain, it may be necessary to use the HTTP POST method, request_uri, or a [Pushed Authorization Request](https://openid.net/specs/openid-federation-1_0.html#RFC9126) [[RFC9126](https://openid.net/specs/openid-federation-1_0.html#RFC9126)] for the request.

The `request_uri` asks the OP to fetch an external resource and this makes it easier to launch DoS attacks against Federation OPs. I propose to say `request_uri` MUST NOT or SHOULD NOT be used by RPs and accepted by OPs.

OIDC recognised this issue and because of that it has the OP `request_uri_parameter_supported` + `require_request_uri_registration` metadata and the RP `request_uris` metadata. So that the `request_uri`'s can be screened before the RP can use them.

In automatic registration if a `request_uri` is used the OP needs to trust it blindly. If the downloaded JWT contains no Trust Chain what should the OP do then? Stop? Or resolve the `client_id`, fetching more resources?

It's good security policy to limit and minimise the number of resources at unverified URLs that need to be fetched. POSTing the `request` object or PAR is sufficient when dealing with large authentication requests (including requests that carry a `trust_chain`).

selfissued:
Fixes #114

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -7091,8 +7091,8 @@ HTTP/1.1 302 Found
           action for the resolve endpoint in this case.
         </t>
 	<t>
-	  Passing request objects by reference using the
-	  <spanx style="verb">request_uri</spanx> request parameter) parameter
+	  Passing request objects by reference (using the
+	  <spanx style="verb">request_uri</spanx> request parameter)
 	  is prohibited in <xref target="authn-request"/> to eliminate a mechanism
 	  by which an attacker could otherwise require OPs
 	  to retrieve arbitrary content under the control of the attacker.
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -5739,7 +5739,8 @@ Content-Type: application/json
 	    by reference
 	    (using the <spanx style="verb">request_uri</spanx> request parameter)
 	    because allowing this would make it easier for attackers
-	    to mount denial of service attacks against OpenID Providers.
+	    to mount denial of service attacks against 
+	    OAuth 2.0 Authorization Servers or OpenID Providers.
 	    They can do this by using the
 	    <spanx style="verb">request_uri_parameter_supported</spanx>
 	    OP metadata parameter with the value <spanx style="verb">false</spanx>.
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -7081,6 +7090,13 @@ HTTP/1.1 302 Found
           the Federation Entity Discovery process should not be the default
           action for the resolve endpoint in this case.
         </t>
+	<t>
+	  Passing request objects by reference using the
+	  <spanx style="verb">request_uri</spanx> request parameter) parameter
```
vdzhuvinov, line 7095: Repeating parameter.

vdzhuvinov (changes requested):
All good, just a typo that needs fixing. Thanks!

rohe (approved):
Provided @vdzhuvinov change request is honoured.

openid-federation-1_0.xml
selfissued: Corrected

vdzhuvinov (approved):

selfissued:
On the 30-Oct-24 Federation editors' call, @ve7jtb suggested the possibility of using HTTP to limit the size of resources received.

Mike said that these OP metadata parameters may be pertinent:
> request_uri_parameter_supported
> OPTIONAL. Boolean value specifying whether the OP supports use of the request_uri parameter, with true indicating support. If omitted, the default value is true.
> require_request_uri_registration
> OPTIONAL. Boolean value specifying whether the OP requires any request_uri values used to be pre-registered using the request_uris registration parameter. Pre-registration is REQUIRED when the value is true. If omitted, the default value is false.

selfissued:
During the 30-Oct-24 Federation editors' call, John was skeptical of removing an OpenID Connect feature.  He suggested ways to mitigate the denial-of-service problems instead, including at the HTTP layer.

peppelinux (approved):

PR #125, opened 2024-10-31
Clarify Trust Mark validation

### LINKED ISSUE #107: Verifying a trust mark
Section 7.3 says:

> To validate a Trust Mark:
> 1. Check the signature of the signed JWT and verify that it has not expired.
> 2. If the Trust Mark identifier appears in the trust_mark_owners claim, verify that the Trust Mark contains a delegation claim. The claims for the Trust Mark identifier in the trust_mark_owners value are used in the following way:
>     - The signature of the delegation JWT MUST verify with a key from jwks claim.
>     - The issuer of the delegation JWT MUST match the sub value in this set of claims.
> 3. Using the Trust Mark issuer status endpoint to verify that the Trust Mark is still active is described in [Section 8.4](https://openid.github.io/federation/main.html#status_endpoint).

For me it's not clear what this list is:
- Is this a list of things an entity could do for validation?
- Is this a list of things an entity must do (in that order) for validation?

For me 1. and 2. seem to belong together, you must do both in order to verify the trust mark.
But 3. falls out (for me).

What is the intended usage of the trust mark status endpoint (keep in mind that it is optional):
- Do entities MUST validate all trust marks there?
- Do entities SHOULD validate all trust marks there?
- Is it an additional thing to do apart from JWT verification?
- Is it an alternative to do JWT verification? (like the resolve endpoint)

Personally, I think the later is the case. But that's not clear from the description.

selfissued:
Fixes #107

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -24,7 +24,7 @@
 
   <front>
     <title abbrev="OpenID Federation">OpenID Federation 1.0 -
-      draft 40
+      draft 41
     </title>
 
     <author fullname="Roland Hedberg" initials="R." role="editor"
@@ -87,7 +87,7 @@
       </address>
     </author>
 
-    <date day="24" month="October" year="2024"/>
+    <date day="30" month="October" year="2024"/>
 
     <workgroup>OpenID Connect Working Group</workgroup>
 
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -3437,8 +3438,11 @@
...
                 </list>
               </t>
+	    </list>
+	  or:
+            <list style="symbols">
               <t>
-                Using the Trust Mark issuer status endpoint to verify that the
+                Use the Trust Mark issuer status endpoint to verify that the
                 Trust Mark is still active
                 is described in <xref target="status_endpoint"/>.
```
zachmann, line 3447:
```suggestion
                as described in <xref target="status_endpoint"/>.
```

zachmann (approved):

peppelinux (approved):
Approved considering Grabiel's suggestion in

vdzhuvinov (approved):

Issue #100, opened 2024-09-30
Trust Anchor Mix-Up (Federation Integrity Property)


SECtim:
Consider the following excerpt from the example of two Federations rooted at two different Trust Anchors as given in Section 1:

```
.-----------------.            .-----------------.
|  Trust Anchor A |            |  Trust Anchor B |
'------.--.-------'            '----.--.--.------'
       |  |                         |  |
    .--'  '---. .-------------------'  |
    |         | |                      |
.---v.  .-----v-v------.   .-----------'
|    |  | Intermediate |   |
'----'  '--.--.--.-----'   |
           |  |  |         |
   .-------'  |  '------.  |
   |          |         |  |
.--v-.      .-v--.     .v--v.
| RP |      |    |     | OP |
'----'      '----'     '----'
```

Now, suppose that `RP` only trusts Trust Anchor `A`, whereas `OP` only trusts Trust Anchor `B`.
A user of `RP` requests to authenticate with `OP`, so `RP` initiates an automatic registration with `OP` (Section 12.1):
1. `RP` uses the process defined in Section 10 to learn `OP`'s metadata, i.e., fetches `OP`'s Entity Configuration and then resolving a Trust Chain, starting with `OP`'s `authority_hints` (i.e., `Intermediate` and `B`). Following the `Intermediate` path, `RP` assembles a Trust Chain from `OP` to `A`, i.e., the chain `[OP, Intermediate, A]`.
2. `RP` validates that Trust Chain and uses it to resolve `OP`'s `openid_provider` metadata, in particular, its authorization endpoint.
3. `RP` sends an authentication request with a signed request object (that does not contain a Trust Chain) to `OP`'s authentication endpoint.
4. `OP` processes this request as described in Section 12.1.1. I.e., among other things, `OP` fetches `RP`'s Entity Configuration and, starting with `RP`'s `authority_hints`, resolves a Trust Chain, say `[RP, Intermediate, B]`.
5. `OP` validates that Trust Chain and uses it to resolve `RP`'s `openid_relying_party` metadata, which `OP` then uses to automatically register `RP`.

I.e., both `OP` and `RP` "trust" each other without having a Trust Anchor in common (actually, `RP` "trusts" `OP` already in Step 3).

Unless we overlooked something, the Federation specification does not prevent this from happening.

Note: Such a situation may also lead to disagreement on the `openid_...` metadata if the different Trust Anchors employ different Federation Policies.

SECtim:
[...]

Issue #135, opened 2024-11-07
Long shot proposal to define additional claim to enable intermediates to limit participation of its leaf entities in different federations


cicnavi:
I'm a bit concerned about the lack of control when it comes to intermediates, its leaf entities and participation in multiple federations. 

Lets say that I have one Intermediate entity `IM1`, which is part of two federations under Trust Anchor `TA1` and Trust Anchor `TA2`. Under my Intermediate `IM1`, I have two subordinates `SUB1` and `SUB2`, and I want to allow `S1` to only participate under `TA1`, and `S2` to only participate under `TA2`. How would I do that?

Solution might be introducing another intermediate entity and dividing subordinates between them. Or, another solution could be introducing extension claim in subordinate statement JWT and marking it critical (adding it under 'crit'), so that it must be understood. Or, another solution could be defining trust marks which designate federation membership like in eduGAIN OIDF PoC project. 

But, I feel that this might be something relatively trivial that could be included in specification out of the box. For example, in Subordinate Statements we could introduce new optional claim like `valid_trust_anchor` (not suggesting the actual name, just trying to make it descriptive) which would designate valid TAs for particular subordinate. 

```
{
  "iss": "https://im1.example.com",
  "sub": "https://sub1.example.com",
  "valid_trust_anchor": [
    "https://ta1.example.com"
  ]  
}
```

With this, trust chain resolution for `SUB1` would only be allowed to end in TAs listed under `valid_trust_anchor` (TA1 in this case). For intermediates regarding subordinate management, this would make it very easy in terms of which subordinate can participate in which federation, using one and only intermediate which is member of multiple federations.

Maybe with this approach, the whole "trust mark architecture" built around eduGAIN OIDF PoC project could be avoided.

cicnavi:
I now realize that this is related to #100, where the possibility of inclusion of "Trust Anchor Hints" in Entity Configurations is mentioned. 

So, in case of introducing TA Hints in Entity Configurations, this proposal would then be the possibility of Intermediate to override / set those values, that is - define valid Trust Anchors in Subordinate Statements for its subordinates.

selfissued:
As I see it, it's not up to subordinates who their superiors have in the trust hierarchy above them.  In fact, that can change dynamically, whether due to mergers and acquisitions, a federation joining an inter-federation, etc.  Allowing subordinates to try to restrict their superior's choices will only result in brittle systems, by sometimes containing obsolete information in the subordinates about who they think their superiors are and what they think they can do.

[...]

PR #138, opened 2024-11-11
feat: added ta_hints for Entity Configurations

peppelinux:
This PR introduces the `ta_hints` as an Entity Configuration parameter, allowing Entities to specify the Trust Anchors they acknowledge and trust.

This addition mitigates the risk of trust anchor mix-ups, particularly when a superior entity is associated with Trust Anchors not recognized by its subordinates.

Trust frameworks may require the use of this parameter to prevent trust evaluations of a Leaf using unauthorized, unknown or rogue Trust Anchors.

This PR is part of the considerations made in the issue #100

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -558,12 +558,30 @@
               the Entities that have at least one Superior above them,
               such as Leaf and Intermediate Entities.
               Its value MUST contain the Entity Identifiers of
-	      its Immediate Superiors and
-	      MUST NOT be the empty array
+	            its Immediate Superiors and
+	            MUST NOT be the empty array
               <spanx style="verb">[]</spanx>.
               This claim MUST NOT be present in Entity Configurations
               of Trust Anchors with no Superiors.
-	      It MUST NOT be present in Subordinate Statements.
+	            It MUST NOT be present in Subordinate Statements.
+            </t>
+            <t hangText="ta_hints" anchor="ta_hints">
+              <vspace/>
+              OPTIONAL. An array of strings representing
+              the Entity Identifiers of known Trust Anchors
+              that the subject signals it trusts.
+              This claim is RECOMMENDED in Entity Configurations of
+              Entities that need to indicate their trusted Trust Anchors.
+              If included, its value MUST NOT be an empty array
+              <spanx style="verb">[]</spanx>.
+              When any values in this parameter match with the
+              Superior Entities listed in the
+              <spanx style="verb">authority_hints</spanx>,
+              it indicates that these Superior Entities are Trust Anchors for
+              this Entity.
+              This claim MUST NOT be present in Entity Configurations
+              of Trust Anchors with no Superiors.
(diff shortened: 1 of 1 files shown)


Razumain:
I recommend to reject this PR.

A TA is trusted to verify other entities (not used by an entity to verify itself). This amendment fails to explain the value of advertising the list of TA this entity is using when trusting other entities.

If there is such value, it should be explained.

peppelinux:
@Razumain I can integrate additional information, even in other parts of the specification, about the value of using this optional parameter

I have deliberately decided to propose this PR in a simple and small way to get further ideas and contributions, like your for which I am grateful to you

peppelinux:
> A TA is trusted to verify other entities (not used by an entity to verify itself). This amendment fails to explain the value of advertising the list of TA this entity is using when trusting other entities.

this PR does not advice about the TA this entity is using when trusting other entities, but the TA the entity recognize above itself.

selfissued:
Having thought about this for a few weeks, I'll repeat my comment https://github.com/openid/federation/issues/135#issuecomment-2502656118, here:
> As I see it, it's not up to subordinates who their superiors have in the trust hierarchy above them. In fact, that can change dynamically, whether due to mergers and acquisitions, a federation joining an inter-federation, etc. Allowing subordinates to try to restrict their superior's choices will only result in brittle systems, by sometimes containing obsolete information in the subordinates about who they think their superiors are and what they think they can do.

I believe this PR should be closed on this basis.

jcmelati:
I have some reservations about this PR and wanted to share my thoughts.

Federations are designed to be dynamic, and adding a `ta_hints` list feels like it could create unnecessary complications. For example, having a `ta_hints` list that’s different from your authority hints is like saying, "I trust these superiors, but I don’t want to be trusted in the context of these superiors." That doesn’t really make sense to me and feels counterintuitive.

Another thing is that federation relationships are naturally bi-directional. Trust isn’t just about one side trusting the other—it’s about both agreeing to trust each other. Adding `ta_hints` seems like it might mess with the natural flow of how these relationships are usually built.

I’m also wondering if `ta_hints` is even necessary. Federation administrators already have ways to set up trust relationships. If an entity only trusts a specific Superior, it makes more sense to connect directly to that Superior. Adding extra fields or lists to manage trust contexts (or worse, invalidating an already established trust with an immediate superior because it’s not listed in `ta_hints`) just seems like overcomplicating things.

There’s also a scalability issue, as @selfissued mentioned. If something changes higher up in the trust chain—like a company getting sold or restructured—it could mean updating `ta_hints` for every entity further down the chain. That kind of ripple effect could stop dependent services until the updates are made, adding a lot of unnecessary work and risking inconsistencies.

That’s just how I see it, though. Maybe I’m missing something here, and I’d love to hear other perspectives to better understand why `ta_hints` might be a good idea.

peppelinux:
we are now fully aware that static trust chains are the best way to provide a verifiable trust path from an entity to a trust anchor

PR #145, opened 2024-11-19
Define additional error codes

### LINKED ISSUE #136: Resolve Endpoint Error Response
The specification does not say much about the error response from the resolve endpoint, only that it follows section 8.9 "Responses from Federation Endpoints", which lists a few different error strings that SHOULD be used.

A resolver might limit its resolving service to only a (sub-)federation and not resolve every trust chain in the world.
In a hierarchical federation, entities therefore might want to query multiple resolvers.
In such a scenario it could make sense to have some error conditions standardized to ease the error handling. 
Especially, it would be helpful to differentiate between the error cases where the resolver does not accept the request (e.g. because it does not support the specified trust anchor) and the case where it cannot find any valid trust chain.

Do you think it is in scope for the specification to specify the error responses in more detail (I would say it's enough to follow the principle of 8.9), or it's entirely up to the implementations?

selfissued:
Fixes #136

Note that we probably also want to harmonize `missing_trust_anchor` with `invalid_trust_anchor` and `trust_chain_validation_failed` with `invalid_trust_chain`.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -6180,7 +6180,8 @@ redirect_uri=https%3A%2F%2Frp.example.com%2Fauthz_cb
 	    as a component of an open redirector.
 	  </t>
 	  <t>
-	    If the OP fails to establish trust with the RP,
+	    If the OP fails to establish trust with the RP or
+	    finds the RP metadata to be invalid or in conflict with metadata policy,
 	    it MUST treat the redirection URI as invalid
 	    and not perform redirection.
 	    This means that the error codes
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -87,7 +87,7 @@
       </address>
     </author>
 
-    <date day="19" month="November" year="2024"/>
+    <date day="26" month="November" year="2024"/>
 
     <workgroup>OpenID Connect Working Group</workgroup>
 
@@ -5173,10 +5173,9 @@ Host: trust-anchor.example.com
 		  The Trust Anchor cannot be found or used.
 		  The HTTP response status code SHOULD be 404 (Not Found).
 		</t>
-		<t hangText="invalid_trust_chain">
+		<t hangText="trust_chain_validation_failed">
 		  <vspace/>
-		  A Trust Chain cannot be constructed or used.
-		  The HTTP response status code SHOULD be 404 (Not Found).
+		  The HTTP response status code SHOULD be 400 (Bad Request).
 		</t>
 		<t hangText="invalid_metadata">
 		  <vspace/>
@@ -6170,7 +6169,9 @@ redirect_uri=https%3A%2F%2Frp.example.com%2Fauthz_cb
 	    If the OP fails to establish trust with the RP,
 	    it MUST treat the redirection URI as invalid
 	    and not perform redirection.
-	    This means that the two error codes defined below,
+	    This means that the error codes
+	    <spanx style="verb">invalid_trust_anchor</spanx> and
+	    <spanx style="verb">trust_chain_validation_failed</spanx>
 	    both of which are about reasons trust failed to be established,
 	    SHOULD only be returned in
 	    <xref target="RFC9126">Pushed Authorization Request</xref>
@@ -6179,20 +6180,9 @@ redirect_uri=https%3A%2F%2Frp.example.com%2Fauthz_cb
           <t>
             In addition to the error codes contained in the
 	    IANA "OAuth Extensions Error Registry" registry <xref target="IANA.OAuth.Parameters"/>,
-	    this specification also defines the following error
-            codes:
-          </t>
-          <t>
-            <list style="hanging">
-              <t hangText="missing_trust_anchor">
-                <vspace/>
-                No trusted Trust Anchor could be found.
-              </t>
-              <t hangText="trust_chain_validation_failed">
-                <vspace/>
-                Trust chain validation failed.
-              </t>
-            </list>
+	    this specification also defines the error codes
+	    in <xref target="error_response"/>,
+	    which MAY also be used.
           </t>
           <figure>
             <preamble>
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -5183,7 +5183,7 @@ Host: trust-anchor.example.com
 		  The Trust Anchor cannot be found or used.
 		  The HTTP response status code SHOULD be 404 (Not Found).
 		</t>
-		<t hangText="trust_chain_validation_failed">
+		<t hangText="invalid_trust_chain">
 		  <vspace/>
 		  The Trust Chain cannot be validated.
 		  The HTTP response status code SHOULD be 400 (Bad Request).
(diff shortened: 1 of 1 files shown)


peppelinux (approved):

openid-federation-1_0.xml
```diff
@@ -5163,11 +5163,30 @@ Host: trust-anchor.example.com
...
+		  <vspace/>
+		  The endpoint cannot service the requested subject.
+		  The HTTP response status code SHOULD be 404 (Not Found).
+		</t>
+		<t hangText="invalid_trust_anchor">
+		  <vspace/>
+		  The Trust Anchor cannot be found or used.
+		  The HTTP response status code SHOULD be 404 (Not Found).
+		</t>
+		<t hangText="trust_chain_validation_failed">
```
vdzhuvinov, line 5181: Missing description - add "The Trust Chain validation failed"?

openid-federation-1_0.xml
```diff
@@ -5163,11 +5163,30 @@ Host: trust-anchor.example.com
...
+		  <vspace/>
+		  The HTTP response status code SHOULD be 400 (Bad Request).
+		</t>
+		<t hangText="invalid_metadata">
```
vdzhuvinov, line 5185: What is the intended use of this error code?

In the spec we say, when the EC `metadata` doesn't pass the `metadata_policy` checks, this must invalidate the Trust Chain (i.e. `trust_chain_validation_failed`).

> If a policy error or another error is encountered during the metadata policy resolution or its application, the Trust Chain MUST be considered invalid.

After that, the OP can take the resolved `metadata` JSON object and try to use to register the RP. So, up to this point, we are in the Federation layer, which is "app layer blind", and the resolved `metadata` is just a JSON object that passed the Trust Chain `metadata_policy` (if any). 

In OAuth / OIDC (dyn reg), when the submitted metadata has an issue, this must result in an `invalid_client_metadata` error.

> invalid_client_metadata
    The value of one of the Client Metadata fields is invalid and the server has rejected this request. Note that an Authorization Server MAY choose to substitute a valid value for any requested parameter of a Client's Metadata. 

Thus, in OIDC, the only use I see for this code is in auto reg, to signal that the Trust Chain validation was successful, but the RP failed to get registered because the resolved `metadata` has an issue. But why not reuse the existing `invalid_client_metadata` code for that?

vdzhuvinov (changes requested):

openid-federation-1_0.xml
zachmann: > In the spec we say, when the EC metadata doesn't pass the metadata_policy checks, this must invalidate the Trust Chain (i.e. trust_chain_validation_failed).

I read this error case as exactly that, but with a more precise hinting, why the validation failed. I think it makes very much a difference if the chain itself is not valid or "only" the metadata. While this still leads to an invalid trust chain, imo a separate error code makes sense.

vdzhuvinov: @zachmann To be specific, you propose the `invalid_metadata` to mean: 

"The Entity metadata does not comply with the resolved metadata policies."

?

openid-federation-1_0.xml
selfissued: Added

openid-federation-1_0.xml
selfissued: I agree with Gabriel that the description of `invalid_metadata` adequately describes what it's for.

vdzhuvinov:
> Please also consider whether we keep the name `trust_chain_validation_failed` or change it to `invalid_trust_chain`. Most of the other error codes are `invalid_SOMETHING`, so this would be more parallel, but it would also be a minor breaking change.

I'd go for `invalid_trust_chain`.

selfissued:
It's now `invalid_trust_chain` for consistency and registration of all the error codes by IANA is now requested.  @vdzhuvinov, can you please re-review?  Thanks.

zachmann (approved):

openid-federation-1_0.xml
```diff
@@ -6158,7 +6180,9 @@ redirect_uri=https%3A%2F%2Frp.example.com%2Fauthz_cb
 	    If the OP fails to establish trust with the RP,
```
vdzhuvinov, line 6180: Please add: 

or finds the RP metadata to be invalid or conflicting metadata policy

openid-federation-1_0.xml
```diff
@@ -6158,7 +6180,9 @@ redirect_uri=https%3A%2F%2Frp.example.com%2Fauthz_cb
...
 	    it MUST treat the redirection URI as invalid
 	    and not perform redirection.
-	    This means that the two error codes defined below,
+	    This means that the error codes
+	    <spanx style="verb">invalid_trust_anchor</spanx> and
+	    <spanx style="verb">invalid_trust_chain</spanx>
 	    both of which are about reasons trust failed to be established,
```
vdzhuvinov, line 6186: The new `invalid_metadata` error falls into the same category. The OP couldn't register the RP metadata because of faulty metadata (or policy conflict).

openid-federation-1_0.xml
```diff
@@ -7420,24 +7436,123 @@ HTTP/1.1 302 Found
           <t>
             <?rfc subcompact="yes"?>
             <list style="symbols">
-              <t>Name: missing_trust_anchor</t>
+              <t>Name: invalid_request</t>
+              <t>Usage Location: authorization endpoint</t>
```
vdzhuvinov, line 7440: This invalid_request error was meant for the federation endpoints, right?

Not sure why we have the authZ endpoint was usage location here.

openid-federation-1_0.xml
```diff
@@ -7420,24 +7436,123 @@ HTTP/1.1 302 Found
...
+              <t>
+                Change Controller: OpenID Foundation Artifact Binding Working Group - openid-specs-ab@lists.openid.net
+              </t>
+              <t>Reference: <xref target="error_response"/> of this specification</t>
+            </list>
+          </t>
+          <t>
+            <list style="symbols">
+              <t>Name: invalid_client</t>
+              <t>Usage Location: authorization endpoint</t>
```
vdzhuvinov, line 7451: Also here.

openid-federation-1_0.xml
```diff
@@ -7420,24 +7436,123 @@ HTTP/1.1 302 Found
...
+              <t>
+                Change Controller: OpenID Foundation Artifact Binding Working Group - openid-specs-ab@lists.openid.net
+              </t>
+              <t>Reference: <xref target="error_response"/> of this specification</t>
+            </list>
+          </t>
+          <t>
+            <list style="symbols">
+              <t>Name: invalid_issuer</t>
+              <t>Usage Location: authorization endpoint</t>
```
vdzhuvinov, line 7462: And here :)

vdzhuvinov (changes requested):
The new `invalid_metadata` error must not allow the `redirect_uri` to be taken.

Noticed that a bunch of the new error codes have the authZ endpoint as the Usage Location.

vdzhuvinov (approved):

Issue #132, opened 2024-11-06
Consistently use Title Case in figures


selfissued:
Let's change "constraints" to "Constraints" in https://openid.net/specs/openid-federation-1_0-40.html#figure-5.

Also, let's change "final Metadata" to "Resolved Metadata" in https://openid.net/specs/openid-federation-1_0-40.html#figure-47.

PR #146, opened 2024-11-20
fix: lables in figure

peppelinux:
This PR resolves https://github.com/openid/federation/issues/132

however, I have to report @selfissued, the term Final Metadata is often used within the draft.
Do we have to change it to Resolved Metadata for consistency?

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -178,7 +178,7 @@
 <!--
   NOTE: this figure was made with ASCIIflow and its editable sources
   are available at the following link:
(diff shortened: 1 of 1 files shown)


selfissued:
"Resolved Metadata" occurs 10 times and "Final Metadata" occurs 3 times - always in examples or figure <name> values.  So I think we should go with "Resolved Metadata" for consistency.

Can you update the PR to also change the other two occurrences of "Final Metadata" to "Resolved Metadata", @peppelinux?  Thanks!

selfissued:
I noticed that this PR didn't update the asciiflow.com link above the example.  Should it have?

And should the link have a right parenthesis ")" at the end of it, or was that a typo?

peppelinux:
@selfissued text aligned to Resolved Metadata

trailing ) is part of the url, not a typo

vdzhuvinov (approved):
Verified the two URLs for the figures - okay!

PR #149, opened 2024-11-27
token_endpoint_auth_method with Automatic Registration

### LINKED ISSUE #133: Possible non-viable outcome for OID DC Registration metadata check in case of Automatic Registration
In OID Federation Automatic Registration, in [12.1.1.1.2. ](https://openid.net/specs/openid-federation-1_0.html#section-12.1.1.1.2)[Processing the Authentication Request](https://openid.net/specs/openid-federation-1_0.html#name-processing-the-authenticati), it says:

> The OP SHOULD furthermore verify that the resolved metadata of the RP complies with the client metadata specification [OpenID Connect Dynamic Client Registration 1.0](https://openid.net/specs/openid-federation-1_0.html#OpenID.Registration) [[OpenID.Registration](https://openid.net/specs/openid-federation-1_0.html#OpenID.Registration)].

OID DC Registration states about the following metadata param: 

> token_endpoint_auth_method
OPTIONAL. Requested Client Authentication method for the Token Endpoint. The options are client_secret_post, client_secret_basic, client_secret_jwt, private_key_jwt, and none, as described in Section 9 of [OpenID Connect Core 1.0](https://openid.net/specs/openid-connect-registration-1_0.html#OpenID.Core) [OpenID.Core]. Other authentication methods MAY be defined by extensions. If omitted, the default is client_secret_basic -- the HTTP Basic Authentication Scheme specified in Section 2.3.1 of [OAuth 2.0](https://openid.net/specs/openid-connect-registration-1_0.html#RFC6749) [RFC6749].

If client has no `token_endpoint_auth_method` metadata defined (it is optional), and if there are no metadata policies defined which would set this metadata value during OID Federation Automatic Registration, according to OID DC Registration we would have to default to `client_secret_basic`, which is not viable since clients won't have client secret.

OID Federation should define default in this case, or add a note in the requirement for "...should verify that metadata complies with OID DC Registration...".

### LINKED ISSUE #151: Implicit assumption of client secret use when automated registration is not used
The third bullet in the section 12.4. Differences between Automatic Registration and Explicit Registration https://openid.net/specs/openid-federation-1_0.html#section-12.4-2.3.1 implies client secret being the only option when automatic registration is not used.

> **Instead of using a Client Secret to authenticate** the client, with Automatic Registration, the client is authenticated by means of the RP proving that it controls a private key corresponding to one of its Entity Configuration's public keys.

I suggest either rewording or removing this bullet point.

selfissued:
Fixes #133 
Fixes #151

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -5752,6 +5752,19 @@ Content-Type: application/json
           <spanx style="verb">client_registration_types_supported</spanx>
           metadata parameter.
 	</t>
+	<t>
+	  Since the RP will have no Client Secret when Automatic Registration
+	  is used, the RP MUST include a
+	  <spanx style="verb">token_endpoint_auth_method</spanx>
+	  parameter in its metadata with a value other than
+	  <spanx style="verb">client_secret_basic</spanx>
+	  and/or include a
+	  <spanx style="verb">token_endpoint_auth_methods_supported</spanx>
+	  parameter in its metadata,
+	  as defined in <xref target="OpenID.RP.Metadata.Choices"/>,
+	  including a value other than
+	  <spanx style="verb">client_secret_basic</spanx>.
+	</t>
 
         <section anchor="authn-request" title="Authentication Request">
           <t>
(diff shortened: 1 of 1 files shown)


rohe (approved):

vdzhuvinov (approved):

cicnavi (approved):

openid-federation-1_0.xml
```diff
@@ -5752,6 +5752,19 @@ Content-Type: application/json
           <spanx style="verb">client_registration_types_supported</spanx>
           metadata parameter.
 	</t>
+	<t>
+	  Since the RP will have no Client Secret when Automatic Registration
+	  is used, the RP MUST include a
+	  <spanx style="verb">token_endpoint_auth_method</spanx>
+	  parameter in its metadata with a value other than
+	  <spanx style="verb">client_secret_basic</spanx>
```
zachmann, line 5760: IMO limiting this to `client_secret_basic` is not sufficient. This would allow the RP to use `client_secret_post` or `client_secret_jwt`, which is equally problematic. The RP would need a value that is no `client_secret_*`.
Same applies for the mentioning of `client_secret_basic` a few lines below.

zachmann (changes requested):

openid-federation-1_0.xml
vdzhuvinov: OAuth and OIDC have no hard bound on the possible client auth methods.

Thus, let's not try to enumerate all possible client auth methods that aren't compatible with auto reg. In 2014 with OIDC we got 4 methods, in 2020 RFC 8705 added 2 more. The OAuth WG is currently working on a draft-ietf-oauth-attestation-based-client-auth draft.

On second reading, the existing paragraph above (line 5742) looks sufficient.

The RP can approach this in ways that are too many to list here:

* It may be subject to a TA policy that resolves to one specific client auth method.
* The resolved OP metadata may include a single `token_endpoint_auth_method`. 
* Etc.

vdzhuvinov (changes requested):
On second reading, I find this addition redundant and unnecessary.

The existing paragraph looks sufficient:

>  Since there is no registration step prior to the Authentication Request, the RP MUST NOT be supplied with a Client Secret.  Instead, the Automatic Registration model requires the RP to use asymmetric cryptography to authenticate its requests.

Maybe reword it like this:

>  Since there is no registration step prior to the Authentication Request, Automatic Registration requires the use of asymmetric cryptography to authenticate its requests. Client Secret-based methods are not compatible.

Common sense. Without trying to enumerate all possible scenarios. They are way too many.

openid-federation-1_0.xml
selfissued: I simplified the treatment, as suggested by @zachmann and @vdzhuvinov.

zachmann (approved):

vdzhuvinov (approved):

vdzhuvinov (approved):

PR #153, opened 2024-11-30
Explain Trust Mark Issuer validation in more detail

### LINKED ISSUE #127: Incomplete statement about validating Trust Mark Issuer
[7.3. ](https://openid.net/specs/openid-federation-1_0-40.html#section-7.3)[Validating a Trust Mark](https://openid.net/specs/openid-federation-1_0-40.html#name-validating-a-trust-mark) includes the statement:
>    "To validate a Trust Mark issuer, follow the procedure defined in [Section 10](https://openid.net/specs/openid-federation-1_0-40.html#resolving_trust)."

where Section 10 is "Resolving the Trust Chain and Metadata".

This seems like an incomplete and confusing statement, because while resolving a trust chain is a necessary step to determine what Trust Anchor to use, that section doesn't describe how to validate a trust mark issuer.

What should we actually say that would be actionable to implementers and where should we say it?

selfissued:
Fixes #127

Note that this PR currently contains four questions prefaced by "TBD" that I'd like people's input on.

This whole section needs to be significantly tightened.  I started by pulling in @rohe's thoughts from the issue, albeit in a rough form.

A rendered version of this PR can be viewed at https://openid.github.io/federation/mbj-trust-mark-issuer-validation.html .

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -3537,7 +3537,6 @@
 	    To determine this, first resolve the Trust Chain
 	    for the Trust Mark Issuer,
 	    as described in <xref target="resolving_trust"/>.
-	    (TBD THEN DO WHAT?)
 	  </t>
       <t>
         Henceforth, "instance" is used as a shorthand for Trust Mark instance. An


openid-federation-1_0.xml
```diff
@@ -3418,14 +3418,35 @@
...
+                as described in <xref target="status_endpoint"/>.
+              </t>
+            </list>
+	    or perform these steps:
             <list style="symbols">
+	      <t>
+		Validate that the Trust Mark Issuer is part of the Federation
+		and that it is possible to get verified metadata about it
+		by using the procedure defined in <xref target="resolving_trust"/>.
+	      </t>
```
zachmann, line 3444: This step would also be required when using the trust mark status endpoint, right? First it must be verified that the TMI is part of the federation and its trust mark status endpoint must be obtained.

openid-federation-1_0.xml
```diff
@@ -3418,14 +3418,35 @@
...
+	      <t>
+		If delegation is not being used (TBD HOW TO DETERMINE THIS?),
```
zachmann, line 3446: Should be fine to check if the `delegation` claim is present in the trust mark.

However, the next points suggests to look at the TA's `trust_mark_owners` claim. So I woudl say there are two approaches:

- TA-centered: First checking if the TM is supposed to use delegation.
- TM-centered: First checking if the TM actually uses delegation.

openid-federation-1_0.xml
```diff
@@ -3436,7 +3457,7 @@
 		of the Trust Anchor's Entity Configuration,
                 verify that the Trust Mark contains
                 a <spanx style="verb">delegation</spanx>
-                claim.
+                claim.  (TBD WHAT TO DO IF IT DOESN'T?)
                 The claims for the Trust Mark identifier
                 in the <spanx style="verb">trust_mark_owners</spanx> value
                 are used in the following way:
```
zachmann, line 3463: For me this does not feel like step 4 of "manual verifying a trust mark": I would prefer to start with the basic things needed; then determining if the trust mark uses (or should use) delegation or not; and then differentiate how a delegated and not-delegated trust mark is verified.

Generally, I think it's helpful to be more clear about delegated trust mark issuance.

openid-federation-1_0.xml
```diff
@@ -3420,14 +3420,35 @@
...
-        <section title="Validating a Trust Mark" anchor="trust_valid">
+        <section title="Validating a Trust Mark" anchor="trust-mark-validation">
           <t>
 	    An Entity SHOULD NOT try to validate a Trust Mark until
 	    it knows which Trust Anchor it is using.
-	    To validate a Trust Mark Issuer, follow the procedure
-	    defined in <xref target="resolving_trust"/>.
+	    To determine this, first resolve the Trust Chain,
+	    as described in <xref target="resolving_trust"/>.
+	    (TBD THEN DO WHAT?)
```
peppelinux, line 3429:
```suggestion
	    To determine this, first resolve the Trust Chain,
	    about the Trust Mark Issuer,
	    as described in <xref target="resolving_trust"/>.
```

rohe (approved):

openid-federation-1_0.xml
```diff
@@ -3447,7 +3468,7 @@
 		of the Trust Anchor's Entity Configuration,
                 verify that the Trust Mark contains
                 a <spanx style="verb">delegation</spanx>
-                claim.
+                claim.  (TBD WHAT TO DO IF IT DOESN'T?)
```
cicnavi, line 3471: Regarding delegation, this might be common sense, but I don't see any instruction for validating that the `trust_mark_id` claim from the Trust Mark JWS is the same as the `trust_mark_id` in Trust Mark Delegation JWS, and that the `iss` from the Trust Mark JWS is the same as the `sub` claim in Trust Mark Delegation JWS. Without this check, Trust Mark Issuer can include `delegation` JWS that is verifiable using `jwks` and `sub` claims from  `trust_mark_owners`, but which actually doesn't say that the Issuer is "allowed" to issue that particular Trust Mark.

openid-federation-1_0.xml
```diff
@@ -3532,9 +3534,11 @@
           <t>
 	    An Entity SHOULD NOT try to validate a Trust Mark until
 	    it knows which Trust Anchor it is using.
-	    To validate a Trust Mark Issuer, follow the procedure
-	    defined in <xref target="resolving_trust"/>.
-          </t>
+	    To determine this, first resolve the Trust Chain
+	    for the Trust Mark Issuer,
+	    as described in <xref target="resolving_trust"/>.
+	    (TBD THEN DO WHAT?)
```
peppelinux, line 3540:
```suggestion
```

peppelinux (approved):

rohe (approved):

PR #188, opened 2025-03-08
fix: naming constraints example

jcmelati:
As per the RFC 5280, `naming constraints` applied to URIs `MUST be specified as a fully qualified domain name and MAY specify a host or a domain. Examples would be "host.example.com" and ".example.com"`.

The current specification incorrectly shows examples with URI scheme prefixes, such as "https://.example.com", which do not conform to RFC 5280.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -3135,10 +3135,10 @@
   "max_path_length": 2,
   "naming_constraints": {
     "permitted": [
-      "https://.example.com"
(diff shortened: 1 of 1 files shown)


selfissued (approved):
As the text is written at https://openid.net/specs/openid-federation-1_0-42.html#name-naming-constraints, @jcmelati is correct.

rohe (approved):

peppelinux (approved):

PR #197, opened 2025-04-14
All claims we list as possible in an Entity Statement should be listed in section 3

rohe:
Whether they are general or not.
If some part of the document says it can be in an Entity Statement then it should be listed in the Entity Statement section.


rohe:
As an implementer I'd like to see all in one place. Not one piece here and another there.

peppelinux (approved):

selfissued (approved):

Issue #199, opened 2025-04-16
Clarify requirement for exp time in Trust Anchor Entity Configuration


jcmelati:
In Section 11.2 of the OpenID Federation 1.0 specification, the following sentence appears:

> A Trust Anchor MUST publish an Entity Configuration about itself. The Trust Anchor SHOULD set an expiration time on its Entity Configuration, such that federation participants will re-fetch it at reasonable intervals.

During an informal discussion before the start of the last AB/Connect WG Pacific call on Monday, April 14, some participants noted confusion arising from the current wording. Specifically, the use of "SHOULD" in relation to the expiration time (exp) might mistakenly suggest that setting an expiration time in the Entity Configuration is optional.

The authors' intention here seems to be to emphasize the best practice of having the Trust Anchor's Entity Configuration re-fetched at "reasonable" intervals. Therefore, the wording must not cause confusion regarding whether the `exp` claim in the Entity Configuration is optional - it is not optional - but rather clarify expectations around the reasonableness of the interval.

There is also ambiguity regarding what constitutes a "reasonable" interval, as different federations will naturally interpret "reasonable" differently. Consequently, the requirement should leave no doubt that an expiration time `exp` MUST be present and that it SHOULD be set to a reasonable interval, defined by the federation's operational requirements.

### Proposed Clarification:

"A Trust Anchor MUST publish an Entity Configuration about itself. The expiration time (exp) set on this Entity Configuration SHOULD ensure that federation participants re-fetch it at reasonable intervals, as determined by the federation's operational requirements."

PR #200, opened 2025-04-16
Clarify requirement for exp time in Trust Anchor Entity Configuration

jcmelati:
PR to address #199

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -5730,11 +5730,10 @@ Content-Type: application/json
       <section title="Key Rollover for a Trust Anchor"
                anchor="key_rollover_anchor">
         <t>
-          A Trust Anchor MUST publish an Entity Configuration about
-          itself. The Trust Anchor SHOULD set an expiration
-          time on its Entity Configuration,
-	  such that federation participants will re-fetch it at reasonable intervals.
-	  When a Trust Anchor rolls over its signing keys, it needs to:
+          A Trust Anchor MUST publish an Entity Configuration about itself. The expiration
+          time (exp) set on this Entity Configuration SHOULD ensure that federation participants
+          re-fetch it at reasonable intervals, as determined by the federation's operational
+          requirements. When a Trust Anchor rolls over its signing keys, it needs to:
         </t>
         <t>
           <list style="numbers">
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -5732,8 +5732,7 @@ Content-Type: application/json
         <t>
           A Trust Anchor MUST publish an Entity Configuration about itself. The expiration
           time (exp) set on this Entity Configuration SHOULD ensure that federation participants
-          re-fetch it at reasonable intervals, as determined by the federation's operational
-          requirements. When a Trust Anchor rolls over its signing keys, it needs to:
+          re-fetch it at reasonable intervals. When a Trust Anchor rolls over its signing keys, it needs to:
         </t>
         <t>
           <list style="numbers">
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -5730,9 +5730,10 @@ Content-Type: application/json
       <section title="Key Rollover for a Trust Anchor"
                anchor="key_rollover_anchor">
         <t>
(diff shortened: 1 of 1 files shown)


teamktown:
Some observations 
- key rollover events for the anchor
  - Keyrollover is good but butter is to have clear statements (MUST's) that the clients need to fetch or check for changes.
    -  otherwise upon key rollover the clients will cease to function as the chain will invalidate due to being signed with a new key
  - Mitigations could be  'sign material/publish with both new and old keys' where clients would walk the N possible keys as equals and if one validates, that satisfies the trust model -- so treat them as a set.  I'm unclear if this is a breaking behaviour change. Operationally it adds resilience, flexibility thus increasing reliability during key change events.  
  - refresh rates/validity duration is a function of tolerance for both risk and of acceptable velocity of key material circulated. 
    - It should be measured in seconds or minutes and not days / hrs
    - Also note that if 'everyone' polls the material at the top of the hour this could create a DDOS effect with many many clients (a flat federation with  anchor and 100's/1000's many leaf nodes. 
    - Mitigation: cache locally and psuedo random fetch period + attempt to validate locally until expiry threshold is met  
 
Fast, resilient, highly available trust circulation matters to keeping TCO as low as possible for all involved.

openid-federation-1_0.xml
```diff
@@ -5730,11 +5730,10 @@ Content-Type: application/json
...
         <t>
-          A Trust Anchor MUST publish an Entity Configuration about
-          itself. The Trust Anchor SHOULD set an expiration
-          time on its Entity Configuration,
-	  such that federation participants will re-fetch it at reasonable intervals.
-	  When a Trust Anchor rolls over its signing keys, it needs to:
+          A Trust Anchor MUST publish an Entity Configuration about itself. The expiration
+          time (exp) set on this Entity Configuration SHOULD ensure that federation participants
+          re-fetch it at reasonable intervals, as determined by the federation's operational
+          requirements. When a Trust Anchor rolls over its signing keys, it needs to:
```
peppelinux, line 5736:
```suggestion
          re-fetch it at reasonable intervals. When a Trust Anchor rolls over its signing keys, it needs to:
```

The Trust Anchor represents the federation and defines the operational requirements, we can say that saying Trust anchor alone is good enough, while including `federation's operational requirements` might introduce some ambiguity or some overlapping elements.

peppelinux (changes requested):
I only suggest to reduce the words as my previous comment

selfissued (changes requested):
"SHOULD ensure" doesn't sound actionable to me.

The existing wording seems equivalent to yours, but seems clearer to me.

What are you trying to achieve with this rewording?

jcmelati:
@selfissued with the rewording I am trying to avoid any confusion that the phrase "The Trust Anchor SHOULD set an expiration time on its Entity Configuration" brings regarding making it appear as the `exp` time attribute on the Entity Configuration is optional.

jcmelati:
@peppelinux merged your correction

openid-federation-1_0.xml
```diff
@@ -5730,11 +5730,9 @@ Content-Type: application/json
...
                anchor="key_rollover_anchor">
         <t>
-          A Trust Anchor MUST publish an Entity Configuration about
-          itself. The Trust Anchor SHOULD set an expiration
-          time on its Entity Configuration,
-	  such that federation participants will re-fetch it at reasonable intervals.
-	  When a Trust Anchor rolls over its signing keys, it needs to:
+          A Trust Anchor MUST publish an Entity Configuration about itself. The expiration
+          time (exp) set on this Entity Configuration SHOULD ensure that federation participants
+          re-fetch it at reasonable intervals. When a Trust Anchor rolls over its signing keys, it needs to:
```
selfissued, line 5735:
```suggestion
          A Trust Anchor MUST publish an Entity Configuration about itself.
          The expiration time (exp) set on this Entity Configuration should be chosen
          such that it ensures that federation participants re-fetch it at reasonable intervals.
          When a Trust Anchor rolls over its signing keys, it needs to:
```
I get it.  Hopefully this slightly updated wording fits the bill.

selfissued (approved):

selfissued:
@peppelinux can you re-review?

peppelinux (approved):

Issue #172, opened 2024-12-24
Add Resolved Metadata to Terminology section


selfissued:
Resolved Metadata is foundational enough to the specification and understanding it that it should be a defined term.

PR #204, opened 2025-04-29
Added Resolved Metadata in the defined terms

peppelinux:
Resolves https://github.com/openid/federation/issues/172

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -381,10 +381,10 @@
             </t>
             <t hangText="Resolved Metadata">
               The metadata that results from applying the metadata policy
-              defined by the Trust Chain to the metadata in the Entity
-              Configuration of the Trust Chain subject. The Resolved Metadata
-              is the metadata that should be used when interacting with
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -379,6 +379,13 @@
               Every participant in a Federation publishes its public
               Federation Entity Keys in its Entity Configuration.
             </t>
+            <t hangText="Resolved Metadata">
+              The metadata that results from applying the metadata policy
+              defined by the Trust Chain to the metadata in the Entity
+              Configuration of the Trust Chain subject. The Resolved Metadata
+              is the metadata that should be used when interacting with
+              the Entity in the context of the Federation.
```
selfissued, line 387:
```suggestion
              The metadata that results from applying the metadata policy
              in the Trust Chain to the metadata in the Entity Configuration 
              for the subject of the Trust Chain. The Resolved Metadata
              is the metadata that is used when interacting with
              the Entity.
```

selfissued (approved):

PR #205, opened 2025-04-29
different signing algorithms for different Entity Statements in a Trust Chains

peppelinux:
Added note about using different signing algorithms for different Entity Statements in a Trust Chain.

@selfissued I would appreciate to add normaitve reference to standards defining the supported algorithm and the security best current practice to exclude the deprecated/weak ones.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -10143,12 +10143,12 @@ Host: op.umu.se
 	  <t>
 	    OAuth 2.0 Protected Resource Metadata is now RFC 9728.
 	  </t>
-      <t>
-        Follow Implementation Considerations in <xref target="OpenID.RP.Choices"/>.
-      </t>
-      <t>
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -492,6 +492,9 @@
           requires support for it (<spanx style="verb">alg</spanx>
           value of <spanx style="verb">RS256</spanx>).
           Federations MAY also specify different mandatory-to-implement algorithms.
+          Note that a Trust Chain can be composed of Entity Statements signed using different
+          signing algorithms, as long as each statement's signature uses a signature algorithm
+          supported by the trust framework in use.
```
selfissued, line 497:
```suggestion
          Note that a Trust Chain can contain Entity Statements signed using different
          signing algorithms, as long as each signature uses a signature algorithm
          supported by the trust framework and implementations in use.
```

selfissued (approved):

PR #217, opened 2025-05-14
Add note about metadata policy and comparing JSON objects

### LINKED ISSUE #35: Using metadata policy on metadata parameters with JSON object values

selfissued:
Fixes #35

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -87,7 +87,7 @@
       </address>
     </author>
 
-    <date day="24" month="April" year="2025"/>
+    <date day="13" month="May" year="2025"/>
 
     <workgroup>OpenID Connect Working Group</workgroup>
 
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -2541,7 +2541,22 @@
                 <spanx style="verb">superset_of</spanx> with identical array
                 values.
               </t>
-
+	      <t>
+		Some JSON libraries may have issues comparing JSON objects.
+		For this reason, support for applying metadata policy to
+		metadata values that are JSON objects is mandatory only for the
+		<spanx style="verb">value</spanx>,
```
zachmann, line 2548: merging `value` operators also requires comparison. Since they can (only) be merged when their values are equal.

peppelinux (approved):

PR #218, opened 2025-05-14
Recommend authenticating RPs to OPs the same way each time

### LINKED ISSUE #147: Client cannot know what client authentication method a server has registered it for
As discussed under https://github.com/openid/federation/issues/12 there is now a mechanism that allows clients to state what options they support.

However there is no way (when using automatic registration at the authorization endpoint) for the client+server to reach agreement on which method will actually be used.

e.g. if the client supports both private_key_jwt and mtls, and the server also supports both, as client that has done automatic registration at the authorization endpoint has no idea if it should send private_key_jwt or mtls to the token endpoint.

Removing automatic registration from the authorization endpoint and hence mandating the use of the PAR endpoint would solve this, as the client would be authenticating to the PAR endpoint and hence the server could register the client with the client authentication method it uses to authenticate at the PAR endpoint. There may be other solutions.

selfissued:
Fixes #147

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -87,7 +87,7 @@
       </address>
     </author>
 
-    <date day="8" month="May" year="2025"/>
+    <date day="13" month="May" year="2025"/>
 
     <workgroup>OpenID Connect Working Group</workgroup>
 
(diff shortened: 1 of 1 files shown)


peppelinux (approved):

selfissued:
Closing for the reasons described in https://github.com/openid/federation/issues/147#issuecomment-2900316207

jogu:
On the WG call today, Mike asked me to comment on whether I thought this would resolve https://github.com/openid/federation/issues/147 or not.

I forget exactly what my response was but I believe it was something along the lines of this getting us partway there - I think we'd need a further statement along the lines:

> In contrast to conventional manual or dynamic client registration, when automatic registration is done at the authorization endpoint it may not be possible for the Authorization Server to determine which client authentication method the client intends to use, as the intersection of the client's supported client authentication methods and the server's may result in more that one algorithm. An authorization server supporting OpenID Federation will hence need to be able to register the client via automatic registration, but (if it needs to assign a single client authentication method to a client) cannot assign a client authentication method to it until the client calls an authenticated endpoint.

PR #224, opened 2025-05-19
Add additional ui related claims for all metadata types.

zachmann:
As we are currently working on a draft for an [Entity Collection Endpoint](https://github.com/zachmann/openid-federation-entity-collection) we found the need for additional metadata claims.

This allows that Entities publish this information in their Entity Configuration and the entity collection endpoint can return the information which then among other use-cases allows to build UIs so End-Users can select an OP.

This PR adds the following claims as **optional** metadata claims that can be used with all entity types:

- display_name
- description
- keywords
- information_uri

This is also inspired by https://docs.oasis-open.org/security/saml/Post2.0/sstc-saml-metadata-ui/v1.0/os/sstc-saml-metadata-ui-v1.0-os.html

I tried to mimic the information specified there, where `policy_uri` and `logo_uri` is already defined in oidfed.

The already existing `organization_name` claim is semantically different from `display_name`, since display name is about the entity and not the organization.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -1769,6 +1769,21 @@
               this MAY be, for example, the person's name. Note
               that this information will be publicly available.
             </t>
+            <t hangText="display_name">
+              <vspace/>
+              OPTIONAL. A human-readable
+              name of the Entity to be presented to the End-User.
+            </t>
+            <t hangText="description">
+              <vspace/>
+              OPTIONAL. A human-readable, brief description of this Entity
+              presentable to the End-User.
+            </t>
+            <t hangText="keywords">
+              <vspace/>
+              OPTIONAL. JSON array with one or more
+              strings representing search keywords, tags, categories, or labels that apply to this Entity.
+            </t>
             <t hangText="contacts">
               <vspace/>
               OPTIONAL. JSON array with one or more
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -1769,6 +1769,21 @@
...
               that this information will be publicly available.
             </t>
+            <t hangText="display_name">
+              <vspace/>
+              OPTIONAL. A human-readable
+              name of the Entity to be presented to the End-User.
+            </t>
+            <t hangText="description">
+              <vspace/>
+              OPTIONAL. A human-readable, brief description of this Entity
```
peppelinux, line 1779: it should be defined with a max length constraint, eg: 1048 chars.

peppelinux (approved):

openid-federation-1_0.xml
```diff
@@ -1787,6 +1802,11 @@
               OPTIONAL. URL of the documentation of
               conditions and policies relevant to this Entity.
             </t>
+            <t hangText="information_uri">
```
selfissued, line 1805: This somewhat overlaps with the meaning of `resource_documentation` at https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#protected-resource-metadata, but as in the previous case, I think this is acceptable.

selfissued (approved):
Thanks for writing this, Gabriel.

openid-federation-1_0.xml
```diff
@@ -1769,6 +1769,21 @@
               this MAY be, for example, the person's name. Note
               that this information will be publicly available.
             </t>
+            <t hangText="display_name">
```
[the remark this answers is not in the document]
zachmann: Yes I'm aware of those claims. I also considered registering defined claims per entity type. But I think this general approach is easier, more straightforward to implement, and also more future-proof with regards to additional entity types that might be defined in the future or by other parties.

PR #225, opened 2025-05-22
Pass Trust Mark to Trust Mark Status Endpoint

### LINKED ISSUE #24: The iat parameter in Trust Mark status requests creates great overhead for no good reason

### LINKED ISSUE #25: Error in Trust Mark status response (8.4.2)
The description of the Trust Mark status response is not logically consistent.

The text states:

>A successful response MUST use the HTTP status code 200 with the content type application/json. The response body is a JSON >object containing the data below:
>
>active
>
>REQUIRED. Boolean. Whether the Trust Mark is active or not.
>
>If the response is negative, the response is as defined in [Section 8.9](https://openid.net/specs/openid-federation->1_0.html#error_response).

This is not consistent.

It is stated that the “active” parameter indicates if the Trust Mark is active or not. This implies that it is valid to respond with a value of “false” if the Trust Mark is not valid.

Then it is stated that a negative response should be provided as defined in section 8.9 (Error responses).

However, the error responses has no means to provide the “active” result. It only returns an error code and a description. And no error code is relevant to a non-active response.

This text should either specify that a negative response (active = false) is a normal response and not according to 8.9. Or it should state that “active” only is used to indicate a successful response if the Trust Mark is valid. The current text makes no sense.

I suggest that a negative response or a positive response are both normal responses (Http status 200) and not according to 8.9. Section 8.9 error responses are only applicable if there is some error in the processing of the request.

This is our code. And I think it is correct:

```java
    boolean valid = hasValidTrustMark(trustMarkId, subject);
    try {
      return new EndPointResponse<>(objectMapper.writeValueAsString(Collections.singletonMap("active", valid)));
    }
```

### LINKED ISSUE #212: Improve clarity of what Trust Mark status means -  And securing the status response
This has been discussed before in various ways, but the problem is still not resolved. I start this issue with a new angle.

This was discussed to length at the workshop in Stockholm and I think that it showed that there is a quite significant ambiguity and disconnect between the standard and how it is implemented.

To explain this issue I start with some definitions for the purpose of this discussion:

**Trust Mark Assignment** - is the approval of a Trust Mark for a particular entity. This also normally have an assignment time and optionally an assignment expiration when this entity will loose the Trust Mark unless the assignment is extended before this time.

**Trust Mark** - is according to the current document the JWT that is provided as a proof of the Trust Mark Assignment.

**Claim**
This issue suggest that when a requestor asks for the status of a TrustMark, they want to know if a subject has a valid Trust Mark Assignment as claimed by this Trust Mark, or whether that Assignment has been revoked. They have no interest in whether the Trust Mark JWT is still valid.

This issue suggest that all implementations worked on in Stockholm, and I would guess other implementations as well, has implemented a status endpoint that delivers "true" if the requested subject has a valid Trust Mark Assignment. One major reason for this is that the Trust Mark endpoint may issue any number of "Trust Marks" for the same assignment, and it then becomes infeasible to respond in the context of a specific JWT. Our implementations do not store issued individual "Trust Mark" JWTs and we keep no history of them.

**Fix**

1. Clarify the language so that the text clearly distinguishes the "Trust Mark" (JWT proof of an assignment) from the Trust Mark Assignment.
2. Clarify that the status endpoint only responds to wether a subject has a TrustMark assignment for the requested "trust_mark_id"
3. Remove "iat" as input parameter as this only refers to a particular Trust Mark JWT.

**Securing the trust mark status response**

Trust mark status response is currently unsigned, which feels unsatisfactory given that this information may be the source of important trust and authorization decisions. There are two ways to solve this:

- Provide a signed response (JWT)
- Remove the Trust Mark status endpoint and let requestors instead just ask for a new Trust Mark. This provides the same information, is already implemented and is already signed.

I personally vote to get rid of the status endpoint altogether. I don't think it adds anything of value over just asking for a new Trust Mark.

selfissued:
This simplifies the Trust Mark Status endpoint by passing the Trust Mark to be validated as a parameter.

Note that this restores functionality that was present until draft 39, including in the last Implementer's Draft (draft 36).

Fixes #24
Fixes #25
Fixes #212

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -87,7 +87,7 @@
       </address>
     </author>
 
-    <date day="21" month="May" year="2025"/>
+    <date day="22" month="May" year="2025"/>
 
     <workgroup>OpenID Connect Working Group</workgroup>
 
@@ -3927,7 +3927,7 @@ Host: edugain.org
             with the content type
             <spanx style="verb">application/entity-statement+jwt</spanx>,
             to make it clear that the response contains an Entity Statement.
-            If it is a negative response, it will be a JSON object and the
+            If it is an error response, it will be a JSON object and the
             content type MUST be
             <spanx style="verb">application/json</spanx>.
 	    If the fetch endpoint cannot provide data for the requested
@@ -4174,7 +4174,7 @@ Host: openid.sunet.se
           </t>
 
           <t>
-            If the response is negative, the response
+            An error response
             is as defined in
             <xref target="error_response"/>.
           </t>
@@ -4401,7 +4401,7 @@ Host: openid.sunet.se
             Additional claims MAY be defined and used in conjunction with the claims above.
           </t>
           <t>
-            If the response is negative, the response
+            An error response
             is as defined in
             <xref target="error_response"/>.
           </t>
@@ -4499,7 +4499,7 @@ Host: openid.sunet.se
 
       <section title="Trust Mark Status" anchor="status_endpoint">
         <t>
-          This enables an Entity to check whether a Trust Mark has been issued to
+          This enables determining whether a Trust Mark has been issued to
           an Entity and is still active. The query MUST be sent to the Trust Mark Issuer.
         </t>
 	<t>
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -4499,7 +4499,7 @@ Host: openid.sunet.se
 
       <section title="Trust Mark Status" anchor="status_endpoint">
         <t>
-          This enables an Entity to check whether a Trust Mark has been issued to
+          This enables determining whether a Trust Mark has been issued to
           an Entity and is still active. The query MUST be sent to the Trust Mark Issuer.
```
zachmann, line 4503:
```suggestion
          This enables determining whether a Trust Mark that has been issued to
          an Entity is still active. The query MUST be sent to the Trust Mark Issuer.
```

zachmann: I think we should make it clear (to prevent any confusion that already arised) that this endpoint is to check if a certain trust mark (jwt) is still valid.

Therefore, I suggest rephrasing this part. The previous statement could be read that this endpoint is used to check if the entity has been issued this trustmark. However, since we already passing the trust mark, that part must be true, but we need to check if the trust mark is **still* valid.

zachmann (approved):

rohe (approved):

Issue #169, opened 2024-12-19
Removed parameters still listed in IANA registry


MichaelFraser1999:
Within the OAuth Authorization Server Metadata section of the [OAuth Parameters Registry](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml), the following parameters are still listed:
- request_authentication_methods_supported
- request_authentication_signing_alg_values_supported

However, they were both removed in version 40. What is the process to get these removed?

selfissued:
When we next go to IANA for registrations for the proposed Final specification, I'll also ask for their removal then.

Note that there are also OAuth error codes that were changed and added.

PR #227, opened 2025-05-27
Rename homepage_uri to organization_uri

selfissued:
Note that an additional change will be needed in the resource metadata registration after #226 is merged.


zachmann (approved):

selfissued:
The set of IANA registrations to be updated in #169 will need to include `homepage_uri` and `organization_uri`.

PR #228, opened 2025-05-28
Define how to validate Entity Statements

### LINKED ISSUE #84: Validation of Explicit Client Registration Response
[Section 12.2.2.2](https://openid.net/specs/openid-federation-1_0-38.html#section-12.2.2.2) describes how an RP validates an OP's response to an explicit client registration request.

Some of the steps are somewhat unclear to me/could be made clearer.

### Valid Entity Statement

> If the response indicates success, the RP MUST verify that its content is _a valid Entity Statement_ and issued by the OP.

What exactly does "verify that its content is a valid Entity Statement" entail  (apart from the issuer check that is mandatory anyway due to "and issued by the OP")?
I suppose the idea is that RP has to check what Section 3 describes, i.e.:
- Validate the signature ("Entity Statement is signed using one of the private keys of the issuer")
- Validate `sub` == RP's Entity ID ("The subject of the JWT is the Entity itself.")
- Validate `typ`
- Validate presence of the required claims, i.e., `iss`, `sub`, `iat`, `exp`, `jwks`

However, this could be made more explicit.
~~Furthermore, my understanding of the current text in Section 12.2.2.2 does not mandate the RP to check the `aud` claim.~~
Edit: The JWT RFC requires checking the `aud` claim (however, this still does not necessarily imply that the `aud` value must be the RP's Entity ID, it just means the the RP can "identify itself" with the `aud` value).

### `authority_hints` leads to Trust Anchor

> The RP MUST verify that at least one of the authority_hints it specified in the Explicit Registration request leads to the Trust Anchor that the OP set in the trust_anchor_id claim.

How exactly does the RP verify this? Is it sufficient to just fetch Entity Configurations? - Probably not. I suppose the intention is for the RP to do a full Trust Chain resolution & validation (which it may of course have already done before sending the registration request, in which case RP may use that cached chain if it has not expired).

This could be clarified by adding "by resolving and validating a corresponding Trust Chain."

### Trust Chain-compliant Metadata

> The RP MUST ensure that the metadata it was registered with at the OP complies with the Trust Chain openid_relying_party policies, which Trust Chain is resolved using the trust_anchor_id and authority_hints claims of the received registration Entity Statement.

This sentence is hard to parse, splitting it into two might help, e.g.,
> The RP MUST ensure that the metadata it was registered with at the OP complies with the Trust Chain openid_relying_party policies. That Trust Chain MUST be resolved using the trust_anchor_id and authority_hints claims of the received registration Entity Statement.

selfissued:
This PR explicitly defines how to validate entity statements.  It begins by defining how to validate all entity statements.

While currently incomplete, I will add rules for validating specific kinds of entity statements, such as those used for Explicit Registration.

Fixes #84

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -87,7 +87,7 @@
       </address>
     </author>
 
-    <date day="26" month="May" year="2025"/>
+    <date day="27" month="May" year="2025"/>
 
     <workgroup>OpenID Connect Working Group</workgroup>
 
@@ -759,6 +759,100 @@
           and use additional claims.
         </t>
 
+	<section anchor="ESValidation" title="Entity Statement Validation">
+	  <t>
+	    Entity Statements MUST be validated in the following manner.
+	    These steps MAY be performed in a different order, provided that
+	    the result - accepting or rejecting the Entity Statement - is the same.
+	    <list style="numbers">
+	      <t>
+		The Entity Statement MUST be a signed JWT.
+	      </t>
+	      <t>
+		The Entity Statement MUST have a
+		<spanx style="verb">typ</spanx> header parameter with the value
+		<spanx style="verb">entity-statement+jwt</spanx>.
+	      </t>
+	      <t>
+		The Entity Statement MUST have an
+		<spanx style="verb">alg</spanx> (algorithm) header parameter
+		with a value that is an acceptable JWS signing algorithm;
+		it MUST NOT be <spanx style="verb">none</spanx>.
+	      </t>
+	      <t>
+		The Entity Identifier of the Entity
+		MUST exactly match the value of the
+		<spanx style="verb">sub</spanx> (subject) Claim.
+	      </t>
+	      <t>
+		The Entity Statement MUST have an
+		<spanx style="verb">iss</spanx> (issuer) Claim
+		with a value that is a valid Entity Identifier.
+	      </t>
+	      <t>
+		Either the <spanx style="verb">iss</spanx> (issuer) Claim value
+		MUST exactly match
+		the <spanx style="verb">sub</spanx> (subject) Claim value,
+		in which case the Entity Statement is this Entity's Entity Configuration,
+		or it MUST exactly match one of the values in the
+		<spanx style="verb">authority_hints</spanx> array,
+		which MUST be present when the <spanx style="verb">iss</spanx>
+		and <spanx style="verb">sub</spanx> Claim values differ
+	      </t>
+	      <t>
+		The current time MUST be after the time represented by the
+		<spanx style="verb">iat</spanx> (issued at) Claim.
+	      </t>
+	      <t>
+		The current time MUST be before the time represented by the
+		<spanx style="verb">exp</spanx> (expiration) Claim.
+	      </t>
+	      <t>
+		The <spanx style="verb">jwks</spanx> (JWK Set) Claim
+		MUST be present, with a value that is a valid
+		JWK Set <xref target="RFC7517"/>.
+	      </t>
+	      <t>
+		If the <spanx style="verb">jwks</spanx> (critical) Claim is present,
+		then for each array element in this claim's value,
+		the Entity Statement MUST contain a claim with the same name
+		as the array element, and that claim MUST be understood
+		and be able to be processed by the implementation.
+	      </t>
+	      <t>
+		Obtain the Entity Configuration for the issuing Entity -
+		the Entity with the Issuer Identifer found in the Entity Statement's
+		<spanx style="verb">iss</spanx> (issuer) Claim.
+		When the <spanx style="verb">iss</spanx> and
+		<spanx style="verb">sub</spanx> Claim values match, this is
+		the Entity Statement being validated itself.
+		Otherwise, this can be obtained either from a Trust Chain or
+		by retrieving it as described in
+		<xref target="federation_configuration"/>.
+	      </t>
+	      <t>
+		The Entity Statement's
+		<spanx style="verb">kid</spanx> (Key ID) header parameter value
+		MUST be a non-zero length string and
+		MUST exactly match the <spanx style="verb">kid</spanx> value
+		for a key in the <spanx style="verb">jwks</spanx> (JWK Set) Claim
+		of the issuing Entity's Entity Configuration.
+	      </t>
+	      <t>
+		The Entity Statement's signature MUST validate using
+		the issuing Entity's key identified by the
+		<spanx style="verb">kid</spanx> value.
+	      </t>
+	    </list>
+	  </t>
+	  <t>
+	    If any of these validation steps fail,
+	    the Entity Statement MUST be rejected.
+	  </t>
+	</section>
+
+	<section anchor="ESClaimsExamples" title="Entity Statement Claims Examples">
+
         <figure>
           <preamble>
             The following is a non-normative example of the JWT Claims Set for an Entity Statement.
@@ -870,6 +964,8 @@
 }
 ]]></artwork>
         </figure>
+
+	</section>
       </section>
 
       <section title="Trust Chain" anchor="trust_chain">
@@ -10316,6 +10412,9 @@ Host: op.umu.se
 	    <spanx style="verb">description</spanx>,
 	    <spanx style="verb">keywords</spanx>, and
 	    <spanx style="verb">information_uri</spanx>.
+	  </t>
+	  <t>
+	    Fixed #84: Added sections on validating Entity Statements.
 	  </t>
 	</list>
       </t>


rohe (approved):

openid-federation-1_0.xml
```diff
@@ -761,9 +761,179 @@
...
+	      </t>
+	      <t>
+		If the <spanx style="verb">trust_marks</spanx> Claim is present,
+		for each element of the array, validate that the
+		<spanx style="verb">trust_mark_type</spanx> member value matches the
+		<spanx style="verb">trust_mark_type</spanx> Claim value in the
+		Trust Mark JWT that is the value of the
+		<spanx style="verb">trust_mark</spanx> member.
+		Validate each Trust Mark as described in
+		<xref target="trust-mark-validation"/>.
```
zachmann: I do not think this should be a requirement for validating entity statements:
I see trust marks as something that a local policy of the validating entity or a federation policy might require on certain entities. i.e. the validator MAY have a list of trust marks it wants to verify. But in my view it does not care about other trust marks that might be present. I do not really see a problem with accepting an entity that publishes an invalid trust mark of type "foobar", if I'm not interested in "foobar" - as long as everything I'm interested in is valid.

I do believe trust mark verification is important, but I believe it is part of establishing trust with an entity not part of verifying an entity statement. Trust mark verification would come after verifying a trust chain.

zachmann: Also according to section 3.2 `trust_marks` can only occur in entity configurations. Other verification steps in this list would verify that the statement is a entity configuration for such claims.

openid-federation-1_0.xml
```diff
@@ -761,9 +761,179 @@
...
+		values that are JSON objects containing
+		a <spanx style="verb">sub</spanx> member with a value that is
+		an Entity Identifier and
+		a <spanx style="verb">jwks</spanx> member with a value that is
+		a JSON Web Key Set.
+	      </t>
+	      <t>
+		If the <spanx style="verb">source_endpoint</spanx> Claim is present,
+		validate that its value is a URL.
+		Implementations MAY also make a fetch call to the the URL
```
zachmann:
```suggestion
		Implementations MAY also make a fetch call to the URL
```

zachmann (commented):
There is some inconsistency with regards to claims that can only occur in entity configurations / subordinate statements. For some claims it is listed as something to check, for others not.

This should be made consistent, either by adding the required check to all claims or by adding a general step to verify that no "wrong" claims are present.

openid-federation-1_0.xml
```diff
@@ -761,9 +761,179 @@
...
+	      </t>
+	      <t>
+		The current time MUST be after the time represented by the
+		<spanx style="verb">iat</spanx> (issued at) Claim
+		(possibly allowing for some small leeway to account for clock skew).
+	      </t>
+	      <t>
+		The current time MUST be before the time represented by the
+		<spanx style="verb">exp</spanx> (expiration) Claim
+		(possibly allowing for some small leeway to account for clock skew).
```
peppelinux: what about moving this to the implementation consideration section and give here a ref to that?

openid-federation-1_0.xml
```diff
@@ -761,9 +761,179 @@
...
+	      </t>
+	      <t>
+		If the <spanx style="verb">trust_mark_owners</spanx> Claim is present,
+		validate that its value is a JSON object with
+		Trust Mark type identifiers as the member names and
+		values that are JSON objects containing
+		a <spanx style="verb">sub</spanx> member with a value that is
+		an Entity Identifier and
+		a <spanx style="verb">jwks</spanx> member with a value that is
+		a JSON Web Key Set.
```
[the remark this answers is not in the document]
selfissued: Ditto

openid-federation-1_0.xml
```diff
@@ -761,9 +761,179 @@
...
+		Trust Mark JWT that is the value of the
+		<spanx style="verb">trust_mark</spanx> member.
+		Validate each Trust Mark as described in
+		<xref target="trust-mark-validation"/>.
+	      </t>
+	      <t>
+		If the <spanx style="verb">trust_mark_issuers</spanx> Claim is present,
+		validate that its value is a JSON object with
+		Trust Mark type identifiers as the member names and
+		arrays of Entity Identifiers as the values.
```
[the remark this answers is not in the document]
selfissued: Done

openid-federation-1_0.xml
```diff
@@ -761,9 +761,179 @@
...
+	      <t>
+		If the <spanx style="verb">crit</spanx> Claim is present,
+		then each array element in this claim's value
+		MUST be a string representing an Entity Statement claim
+		that is not defined by this specification
+		and that claim MUST be understood
+		and be able to be processed by the implementation.
+	      </t>
+	      <t>
+		If the <spanx style="verb">metadata</spanx> Claim is present,
```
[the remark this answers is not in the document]
selfissued: Will be addressed by #275

openid-federation-1_0.xml
```diff
@@ -761,9 +761,179 @@
...
+		Trust Mark type identifiers as the member names and
+		values that are JSON objects containing
+		a <spanx style="verb">sub</spanx> member with a value that is
+		an Entity Identifier and
+		a <spanx style="verb">jwks</spanx> member with a value that is
+		a JSON Web Key Set.
+	      </t>
+	      <t>
+		If the <spanx style="verb">source_endpoint</spanx> Claim is present,
+		validate that its value is a URL.
```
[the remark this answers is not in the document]
selfissued: Done

openid-federation-1_0.xml
zachmann: I'm fine with requiring validating correct syntax. Requiring to verify that the trust mark types in the object match the type in the JWT requires decoding of the JWT which is some effort. But I could also live with that.

But I'm against requiring to verify all trust marks, this means significant effort. I'm fine with just removing the last sentence `Validate each Trust Mark as described in <xref target="trust-mark-validation"/>.`

peppelinux (approved):

openid-federation-1_0.xml
```diff
@@ -761,9 +761,193 @@
...
+		<xref target="federation_configuration"/>.
+	      </t>
+	      <t>
+		The Entity Statement's
+		<spanx style="verb">kid</spanx> (Key ID) header parameter value
+		MUST be a non-zero length string and
+		MUST exactly match the <spanx style="verb">kid</spanx> value
+		for a key in the <spanx style="verb">jwks</spanx> (JWK Set) Claim
+		of the Entity Configuration of the issuing Entity.
+	      </t>
```
teamktown: To better support key rollover, replace 837 to 843 with:

The Entity Statement's <spanx style="verb">kid</spanx> (Key ID) header parameter value
MUST be a non-zero length string and
MUST exactly match the <spanx style="verb">kid</spanx> value
for a key in the <spanx style="verb">jwks</spanx> (JWK Set) Claim
of the Entity Configuration of the issuing Entity.
The <spanx style="verb">jwks</spanx> Claim MAY contain multiple active signing keys
(for example, during key rotation). A validator MUST accept any signature
that verifies successfully with a currently published key in that set.
Publishers SHOULD retain the retiring key alongside the new key
for an overlap period sufficient to allow dependent Entity Statements
to refresh before the old key is withdrawn.

openid-federation-1_0.xml
```diff
@@ -761,9 +761,193 @@
...
+	      <t>
+		If the <spanx style="verb">trust_anchor</spanx> Claim is present,
+		validate that its value is an Entity Identifier.
+		Implementations MAY retrieve the Entity Configuration for the
+		Entity Identifier and validate that it is a Trust Anchor,
+		and they MAY also validate that it is the Trust Anchor
+		used for the Explicit Registration.
+		This Claim MUST NOT be present in Entity Statements that are not
+		Explicit Registration responses.
+	      </t>
```
teamktown: update 932-939 with this to support possible multiple trust anchors 
If the <spanx style="verb">trust_anchor</spanx> Claim is present,
validate that its value is an Entity Identifier corresponding to one
of the configured Trust Anchors in the validator’s Anchor Set.
Implementations MAY retrieve the Entity Configuration for the
Entity Identifier and validate that it is a Trust Anchor,
and they MAY also validate that it is the Trust Anchor
used for the Explicit Registration.
Validators MAY also accept Explicit Registration responses that omit
the <spanx style="verb">trust_anchor</spanx> Claim if the associated Trust Chain
terminates at any Trust Anchor in the Anchor Set.

teamktown (commented):
- [x] Augmented some existing comments alongside others and the flow of trustmark verification.
- [x] added where necessary the recognition of multiple keys for CA rollover into the Entity Statement verification process so that it allowed for a set of CAs
- [x] add some operational aspects to key rollover section

No new fields, no behaviour removals. Only action item is to ensure the validator picks the correct key by kid from a set and to optionally allow multiple anchors. No durations mandated and included guidance to model grace window to reflect refresh expectations to not be in conflict thus avoiding to short a refresh window for slow refreshes thus breaking trust due to rollover anchor aged out but not yet fetched.

While back compatible with deployed OpenIDFed it will reveal where libraries are overly rigid to a single key/single TA and not a set. There are clear, testable paths forward with this that in turn will mean more seamless keyrollover experience.

openid-federation-1_0.xml
selfissued: The validation section isn't the place to add statements about how the "jwks" claim is used.  If you want to file an issue suggesting text to add in the claim definition, I'd welcome that.

openid-federation-1_0.xml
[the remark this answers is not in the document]
selfissued: This is another comment that probably warrants its own issue.  Please file one with suggested wording updates.  Thanks!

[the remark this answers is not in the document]
teamktown: Got it and accept the recommendation as in previous comments. Will work through the way to get these items organized for discussion.

PR #231, opened 2025-05-31
Align terminology in diagram with section header

jmandel:
When Figure 47 says "Evaluate the Trust Chain", I think it's referring to the process described in https://openid.net/specs/openid-federation-1_0.html#section-10.2 which is called "Validating a Trust Chain". If so, it would be good to use the same wording ("Validate" rather than "Evaluate") in the Figure.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -5628,7 +5628,7 @@ Content-Type: application/json
    |                               |        Subordinate Statement about RP  |
    |                               |<---------------------------------------|
    |                               |                                        |
-   |                               | Evaluates the Trust Chain              |
+   |                               | Validates the Trust Chain              |
    |                               |--------------------------              |
    |                               |                         |              |
    |                               |<-------------------------              |


selfissued:
There are 10 places that the word "evaluate" (including "reevaluate") occurs in the specification, including another occurrence in the same figure.  I'm not necessarily opposed to the proposed change, but if we're going to do it for this one occurrence, we should consider what wording we want to use in the other occurrences as well, so that we're being consistent overall.

jmandel:
Agree -- evaluate, resolve, and validate would all be good to define or narrow down if there's redundancy

selfissued:
Josh, we discussed this on the Federation editors' call today.  We agree with Roland that what's going in this step is more than validation - it includes using the data in the trust chain.  Therefore, we're not going to apply the suggested change as is.

That said, if you want to suggest wording updates to the text in the section [10. ](https://openid.net/specs/openid-federation-1_0.html#section-10)[Resolving the Trust Chain and Metadata](https://openid.net/specs/openid-federation-1_0.html#name-resolving-the-trust-chain-a) to better describe this, we'd welcome suggestions.  Thanks.

PR #232, opened 2025-06-29
Describe Client Authentication with Automatic Registration

### LINKED ISSUE #147: Client cannot know what client authentication method a server has registered it for

selfissued:
Fixes #147

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -6350,7 +6350,7 @@ HTTP/1.1 302 Found
 	  </t>
 	  <t>
 	    OPs SHOULD accept any client authentication method that is mutually supported
-	    and RPs SHOULD only use mutually supported methods.
+	    and RPs MUST only use mutually supported methods.
 	    Because some OPs may be coded in such a way that
 	    they expect the RP to always the same client authentication method
 	    for subsequent interactions, note that
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -76,7 +76,7 @@
       </address>
     </author>
 
-    <date day="28" month="June" year="2025"/>
+    <date day="30" month="June" year="2025"/>
 
     <workgroup>OpenID Connect Working Group</workgroup>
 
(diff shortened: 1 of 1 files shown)


jogu (approved):

jogu (changes requested):
Initially I thought this was sufficient, but Tim's comment ( https://github.com/openid/federation/issues/147#issuecomment-3034477543 ) has caused me to consider.

https://github.com/openid/rp-metadata-choices doesn't define multi-values-capable entries for `revocation_endpoint_auth_methods_supported` or `introspection_endpoint_auth_methods_supported`. I think we need to either:

1. Define these, or:
2. Make clear that  the token endpoint auth methods is used for all endpoints

selfissued:
@jogu, does https://github.com/openid/rp-metadata-choices/pull/7 do the trick so that you can approve this PR?

rohe:
It would be interesting to know if the OIDC test suite allows a client to use different client authentication methods at different endpoints. The OIDC standard allows it and therefor the test suite should be able to support it.

bc-pi:
> The inclusion of `revocation_endpoint_auth_methods_supported` and `introspection_endpoint_auth_methods_supported` in AS Metadata (RFC8414) remains a mystery to me

Same but it shouldn't be a mystery that it was a mistake. 

> but I do recall a distant conversation with @b---c that we shall not be registering any more of those endpoint-specific AS metadata for new endpoints. PAR, CIBA, Device Authorization Grant in fact all state the client authenticates in the same manner as for the token endpoint.

I recall this 

> 
> > I think we need to either:
> > 
> > 1. Define these, or:

Absolutely not. 

> > 2. Make clear that the token endpoint auth methods is used for all endpoints
> 
>     2. is right, we do not have a general client data model that supports the client having different auth methods at different endpoints. We do not need one. And I would even go as far as to say the concern of a client out of the blue using different auth methods for different endpoints is largely theoretical, not practical.

Agree that 2 is right.

selfissued:
@jogu, in light of https://github.com/openid/rp-metadata-choices/pull/8 and the clarification to use `token_endpoint_auth_methods_supported` for all AS endpoints, can you please re-review and consider approving?  Thanks.

PR #234, opened 2025-07-11
Fixes syntax errors and figure 5

eduperottoni:
- Replaces `trust_anchor_id` for `trust_anchor`
- Replaces `entity_id` for `Entity Identifier`
- Changes figure 5: `constraints` _claim_ must be only on Subordinate Statement


peppelinux (approved):

openid-federation-1_0.xml
```diff
@@ -12063,6 +12063,7 @@ Host: op.umu.se
 	Tim Würtele,
 	Kristina Yasuda,
 	Gabriel Zachmann,
+	Eduardo Perottoni,
```
selfissued, line 12066: Please move your name to be in alphabetical order by last name.  That would place you between Alexey Melnikov and Roberto Polli.

eduperottoni: Oh, I didn't realize that was the order. Fixed!

eduperottoni:
> @peppelinux, can you update the asciiflow source in line 1021 to match the new diagram contents and put it in the comment thread here so @eduperottoni can add it to the PR? Thanks.

I've just uploaded the URL of the updated ASCIIFlow image :)

rohe (approved):

PR #239, opened 2025-08-06
Changed the value of the trust_chain parameter into something parsable.

rohe:
The old value was invalid. Signatures could not be verified.


ubamrein:
Checked with our implementation, seems to work :)

ubamrein:
Ahm so the header of the JWTs

(e.g. 
`eyJhbGciOiJSUzI1NiIsImtpZCI6IlNDMXBkbXhuUlRJNWFUSlRkRXQ0WjJaWVJtc3lObWhZTTI1TlQzTTVVM1YyTlVKUVptTTNaR3hIVFEifQ`

```json
{
  "alg": "RS256",
  "kid": "SC1pdmxnRTI5aTJTdEt4Z2ZYRmsyNmhYM25NT3M5U3V2NUJQZmM3ZGxHTQ"
}
```
)

Don't contain a `typ` claim. But Section 3 https://openid.net/specs/openid-federation-1_0.html#name-entity-statement mandates one. Shouldn't that be adjusted as well?

Issue #245, opened 2025-09-02
Failing on chain loops


Razumain:
Section 10.1 states:

> Federation participants MUST NOT attempt to fetch Entity Statements they already have obtained during this process to prevent loops. An invalid_trust_chain error SHOULD be returned if a loop is encountered.

I believe this statement is incorrect.

I think the first part is correct, but as long as you are able to build a valid path, it should be OK. I fail to see why you "SHOULD" fail.

Razumain:
I think you miss my point. The fact that you "encounter" a loop, does not mean that you incorporate it in your validated chain.

Do you see the difference?

Say there is a valid chain:  
A -> B -> C -> D -> E

Say that this chain also has a potential loop that you could be tricked into if you are not careful:
A -> B -> C -> D -> B -> C -> D -> B -> C -> D -> B -> C -> D -> B -> C -> D ....

Just because you "encounter" = observe this loop, does not make the valid chain invalid. (A -> B -> C -> D -> E)

Maybe this is just a wording thing. As long as you find a valid chain that has no repetition (loops) in it. You are good and should not fail because it is possible to build a loop.

selfissued:
Addressed by #253

Razumain:
@selfissued I made a comment to the PR that you didn't respond to. Was that because you didn't see it, or because you dismissed it?

PR #253, opened 2025-09-25
If a loop is detected while fetching a trust chain, that chain should be ignored.

rohe:
The whole process of fetching trust chains should not fail because one chain ends in a loop.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -5572,9 +5572,7 @@ Content-Type: application/json
         <t>
           Federation participants MUST NOT attempt to fetch
           Entity Statements they already have obtained during this
-          process to prevent loops.
-	  An <spanx style="verb">invalid_trust_chain</spanx> error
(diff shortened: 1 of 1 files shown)


fkj:
Fixes #245.

openid-federation-1_0.xml
```diff
@@ -5572,9 +5572,7 @@ Content-Type: application/json
         <t>
           Federation participants MUST NOT attempt to fetch
           Entity Statements they already have obtained during this
-          process to prevent loops.
-	  An <spanx style="verb">invalid_trust_chain</spanx> error
-	  SHOULD be returned if a loop is encountered.
+          process to prevent loops. If a loop is detected the authority hint that lead to it MUST be ignored.
```
selfissued, line 5575:
```suggestion
          process to prevent loops.
          If a loop is detected, the authority hint that led to it MUST NOT be used.
```

selfissued (approved):

Issue #249, opened 2025-09-14
Guidance on how to respond to unknown trust mark on status endpoint


MichaelFraser1999:
The trust mark status endpoint states that

`The query MUST be sent to the Trust Mark Issuer.`

And further down, we have 

`An error response is as defined in [Section 8.9](https://openid.net/specs/openid-federation-1_0.html#error_response).`.

It isn't overly clear which error response a server is expected to return in the event they receive a request for the status of a trust mark it did not issue or is not aware of. I can see implementations currently picking one of `invalid_request` or `not_found` (though `not_found` does explicitly scope itself to an unknown Entity Identifier right now so perhaps not this one).

I'd like to know others thoughts on this but I would suggest we define a dedicated `invalid_trust_mark` error code for this use case and add guidance text to use it in the trust mark status response.

zachmann:
I would not expect an error response at all, but instead an `{"active": false}` similar to RFC7662.

MichaelFraser99:
That said... I still don't think an `{"active": false}` response is correct when presented with a trust mark that the server didn't issue - i.e. an explicitly incorrect `iss` value within. Returning as such could lead to greater confusion

rohe:
So perhaps returning the HTTP status code 404 would be more appropriate.

PR #256, opened 2025-10-02
Requested status for an unknown trust mark.

rohe:
We need to define how a server must respond if it receives a request for the status of a trust mark it did not issue or is not aware of.

Fixes https://github.com/openid/federation/issues/249

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -4605,7 +4605,7 @@ trust_mark=eyJ0eXAiOiJ0cnVzdC1tYXJrK2p3dCIsImFsZyI6 ...
             <xref target="error_response"/>.
           </t>
             <t>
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -4604,7 +4604,10 @@ trust_mark=eyJ0eXAiOiJ0cnVzdC1tYXJrK2p3dCIsImFsZyI6 ...
             is as defined in
             <xref target="error_response"/>.
           </t>
-
+            <t>
+                If the server receive a request for the status of a trust mark it did not issue or is not aware of,
```
peppelinux, line 4608:
```suggestion
                If the Trust Mark Issuer receives a request about the status of an unknown Trust Mark, something it did not issue or is not aware of,
```

peppelinux (approved):

zachmann (approved):

PR #265, opened 2025-10-11
feat: added discovery patterns section

peppelinux:
This PR introduces a dedicated section about the discovery patterns implementers should implement according to specific use cases.

This PR aims to describe the different ways to implement trust discovery according to different approaches. These approaches were collected during the past few years and based on real-world implementations and implementers' experiences.

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -78,7 +78,7 @@
       </address>
     </author>
 
-    <date day="10" month="November" year="2025"/>
+    <date day="17" month="November" year="2025"/>
 
     <workgroup>OpenID Connect Working Group</workgroup>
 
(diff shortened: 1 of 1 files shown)


openid-federation-1_0.xml
```diff
@@ -10611,6 +10709,9 @@ Host: op.umu.se
 	    Fixed #147: Added a note about client authentication methods
 	    and Automatic Registration.
 	  </t>
+	  <t>
+	    Added Discovery Patterns section.
+	  </t>
```
selfissued: Please merge with main and then move the history entry into the -45 section.

openid-federation-1_0.xml
```diff
@@ -7719,6 +7721,169 @@ HTTP/1.1 302 Found
...
+
+	<t>
+	  Implementers may support one or more of the following patterns:
+	  <list style="symbols">
+	    <t>
+	      <xref target="top_down_discovery">Top-Down Discovery</xref>: Finding entities within a federation
+	      by starting from a Trust Anchor and traversing down the hierarchy.
+	    </t>
+	    <t>
+	      <xref target="bottom_up_discovery">Bottom-Up Trust Chain Resolution</xref>: Resolving Trust Chains
```
selfissued: I suggest reordering this list and the following subsections to place bottom up discovery first in the descriptions, as that is the primary discovery mechanism defined by the specification.  We should lead with it.  Thanks.

selfissued (approved):
I reordered the sections to put Bottom-Up Trust Chain Resolution first, with @peppelinux's permission.

Issue #216, opened 2025-05-13
Make it optional to provide an actual Trust Mark in Entity Configuration, or allow expired Trust Marks


Razumain:
I forgot about this one, also discussed in Stockholm. I include it here for completeness.

Many implementers are considering short lived Trust Mark JWT:s.

At the same time we want leaf entities to publish Entity Configuration that includes all Trust Marks. This makes Entity Configuration publication more dynamic than it should need to be.
This is the only thing that prevents the Entity Configuration to be more long lived and updated only in case of information update (or expiry).
This also requires the leaf entity to implement the Trust Mark endpoint API for continuous update of Trust Marks.

It would be helpful if inclusion of the actual Trust Mark is optional. If the Trust Mark is absent, the verifier simply has to request a new one.

The downside of this is that the verifier would need to know the id of the issuer to do this, and that would require a change to the current format.

An alternative is to keep the spec as it is, but clearly state that it is legal to provide an old expired Truts Mark. It is then up to the verifier to request a new one from the same issuer.

In either case, I think some clarifying language would be helpful.

selfissued:
@zachmann, if you have the Trust Mark JWT, you don't also need a `trust_mark_id` parameter because you have the `trust_mark_id` in the JWT itself.

@Razumain, it's already optional to provide Trust Marks in Entity Configurations.  Not all Entities need them.  As for expired Trust Marks, they should be disregarded.

@Razumain, it sounds like what you're actually asking for is a new feature that if you have a `trust_mark_id` claim and possibly also a `trust_mark_issuer` claim in an Entity Configuration, then that's interpreted as a signal that a Trust Mark with that ID can be obtained from that Issuer when needed.  Do I have that right?

zachmann:
> [@zachmann](https://github.com/zachmann), if you have the Trust Mark JWT, you don't also need a `trust_mark_id` parameter because you have the `trust_mark_id` in the JWT itself.

While this is conceptual true, the spec (currently) REQUIRES that the `trust_mark_id` (as well as the `trust_mark_jwt`) is given. And this for good reason. The `trust_mark_id` is used to find the correct trust mark without the need to actually inspecting it.

I think I'm on the same page here with Stefan.

zachmann:
> If you don't have the trust_mark how would you know which issuer to ask for a new trust_mark ??

As noted above:
[...]

PR #266, opened 2025-10-12
Added two more implementation considerations.

rohe:
Experience from designing OpenID federations has show that there are some considerations that are worth thinking about.

Fixes https://github.com/openid/federation/issues/216


openid-federation-1_0.xml
```diff
@@ -7386,6 +7386,21 @@ HTTP/1.1 302 Found
           as mandated in <xref target="fetching-es"/>.
 	</t>
       </section>
+        <section anchor="anchor_resolver" title="Trust Anchors and Resolvers go together">
+            <t>
+            If you plan to have one resolver in your federation you should have the same entity being
+            both the Trust Anchor and the Resolver. If you do that users of the resolver will not have to
+            collect and evaluate a trust chains for the Resolver since you must trust the Trust Anchor and by
+            combining the two services the Resolver will be implicitly trusted since the Trust Anchor is trusted.
```
selfissued:
```suggestion
            If you plan to have one resolver in your federation, you should have the same entity be
            both the Trust Anchor and the Resolver. If you do that, users of the resolver will not have to
            collect and evaluate Trust Chains for the Resolver, since you must trust the Trust Anchor. By
            combining the two services, the Resolver will be implicitly trusted since the Trust Anchor is trusted.
```

openid-federation-1_0.xml
```diff
@@ -7386,6 +7386,21 @@ HTTP/1.1 302 Found
...
+            </t>
+    </section>
+        <section anchor="one_service" title="One entity one service">
+            <t>
+            Apart from letting an entity provide both the Trust Anchor and Resolver services there is a good reason for
+            letting each entity only do one thing. The reason is that, later in time, it will be much easier to share specific services
```
selfissued:
```suggestion
            Apart from letting an entity provide both the Trust Anchor and Resolver services, there is a good reason for
            having each entity only do one thing. The reason is that, later in time, it will be much easier to share specific services
```

selfissued (approved):
Approved.  Please apply my requested changes.

peppelinux (approved):

rohe:
Such a mess. Will start over.

PR #274, opened 2025-10-29
Examples using Automatic Registration at other OAuth 2.0 endpoints

### LINKED ISSUE #233: Trust Chain Resolution for calls to Introspection Endpoint by a Federated Protected Resource
Hello,

I'm seeking clarification on the intended interaction between the OpenID Federation trust model and the OAuth 2.0 Token Introspection endpoint. The specification is clear about trust establishment during the authorization flow, but less explicit about subsequent interactions.

**Scenario:**

1. A Client is registered with an OpenID Provider (OP).
2. A Protected Resource (PR) exists, which is not pre-registered with the OP, but shares a common Trust Anchor with it.
3. The Client obtains an access token from the OP and presents it to the PR.
4. The PR now needs to validate this access token by calling the OP's introspection endpoint.

According to [RFC 7662](https://datatracker.ietf.org/doc/html/rfc7662), to prevent token scanning attacks, the introspection endpoint should be protected and require the client (in this case, the PR) to authenticate itself. Since the PR is not registered, it cannot use a pre-configured `client_id` or `client_secret`.

When the OP receives this authenticated call from an unknown entity (the PR), is the following interpretation correct?

1. The OP must first perform Trust Chain Resolution on the identity of the calling PR.
2. If a valid trust chain can be built to a shared Trust Anchor, the OP can consider the PR authenticated and trusted. It can then proceed to introspect the access token and return a response.
3. If no valid trust chain can be established, the OP should reject the request with an authentication error, even if the presented access token itself might have been valid.

This contrasts with a call to the Userinfo endpoint, where the authorization is based on the bearer token itself, and no separate trust establishment for the calling PR seems necessary.

Could you confirm if this interpretation is correct? If so, would you consider adding a note or an example to the specification clarifying this? It would provide valuable guidance for implementers on how to correctly secure endpoints in a federated environment.

selfissued:
Fixed #233

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -78,7 +78,7 @@
       </address>
     </author>
 
-    <date day="28" month="October" year="2025"/>
+    <date day="3" month="November" year="2025"/>
 
     <workgroup>OpenID Connect Working Group</workgroup>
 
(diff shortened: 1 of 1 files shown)


eduperottoni:
> [eduperottoni](https://github.com/eduperottoni) can you please review?

@selfissued I'm sorry... I don't know why, but I dind't receive the notification about this message and that's why I didn't answer here. It's good for me the way you did it. 👍🏽

Issue #269, opened 2025-10-16
Editorial: Restructure 7.3 for clarity. Make trust mark issuer validation requirement more explicit.


lj-raidiam:
The second sentence contains a parenthetical reference to Trust Mark Issuer validation that is somewhat out of the main context but that should be a mandatory prerequisite. 

I propose to make it more explicit mandatory requirement that establishing trust in the issuer must occur before validating the Trust Mark itself. 

Something along these lines:
An Entity MUST establish trust in the Trust Mark Issuer by following the procedure defined in the Section 10 prior to starting the Trust Mark validation process defined below.

Issue #270, opened 2025-10-16
Editorial: 7.3 Entity -> Trust Mark Issuer


lj-raidiam:
In the [7.3](https://openid.net/specs/openid-federation-1_0.html#section-7.3) under the bullet points we use the term _... Entity representing the accreditation authority ..._

I propose to be more direct and change the Entity to Trust Mark Issuer or add a Trust Mark Issuer to that sentence for clarity. That's especially that we this paragraph is about inclusion of this entity in the `trust_mark_issuers` claim.

TomCJones:
the trust mark issuer does not need to be in the federation

PR #276, opened 2025-11-03
Trust mark validation

rohe:
Validating a Trust Mark is separate from using it.

A federation may have a policy that says that only trust marks issued by Trust Mark Issuers that belongs to the federation but that is separate from validating a specific Trust Mark.

Fixed https://github.com/openid/federation/issues/269, https://github.com/openid/federation/issues/270

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -7733,8 +7733,8 @@ HTTP/1.1 302 Found
                 <t hangText="Validating Trust Marks in the Context of Validating an Entity Statement">
             <vspace/>
                     According to the text on Entity Statement Validation in <xref target="ESValidation"/>, validating a
-                    Trust Mark is confined to validating the syntax of the claim value and verifying the
-                    <spanx style="verb">trust_mark_type</spanx> connection.
+                    Trust Mark is confined to validating the syntax of the claim value,
+                    including that the <spanx style="verb">trust_mark_type</spanx> value is consistent.
                 </t>
                 <t hangText="Validating a Specific Trust Mark">
                     <vspace/>


openid-federation-1_0.xml
```diff
@@ -3629,19 +3628,6 @@
...
-            appear for various reasons, such as the Entity Configuration including Trust Marks
-            associated with another federation, or Trust Marks intended for specific
-            purposes or Entity audiences.
-          </t>
-          <t>
-            An Entity MAY choose, at its own discretion, to utilize Trust Marks presented
-            to it that are not recognized within the federation, and where the
-            accreditation authority is established by an out-of-band mechanism.
-            In that case, validation Step 1 becomes OPTIONAL.
-          </t>
```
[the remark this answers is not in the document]
rohe: In my mind there are three things involved here:
1) Validating trust marks in the context of validating an Entity Statement
According to the latest text on Entity Statement Validation validating a Trust Mark is confined to validating the syntax of the claim value and verifying the trust_mark_type connection.
2) Validating a specific Trust Mark
This is what is described in section 7.3.
In order to validate a Trust Mark the entity must find a trust chain for the Trust Mark Issuer to a Trust Anchor the entity trusts. This has nothing to do with which federation that will later be used for the application protocol.
3) Deciding which Trust Marks to use
There might be a federation policy for a federation that states that only trust marks that are issued by trust mark issuers in that federation can be used but that is a policy for the usage of the trust marks. It is something applied after the trust mark has been validated.

Given this division of labors I removed these two paragraphs since they concerned the third bullet and not the second.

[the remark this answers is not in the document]
zachmann: I kind of agree and kind of disagree. I fully agree with Mike that those point must be address somewhere.

I'm fine with moving this to some other place, but we then get a problem with bullet point 1 of the verification steps, that MUST all be done:

> Verify that the value of the instance’s trust_mark_type is listed in the Trust Anchor's trust_mark_issuers, and if so, that the instance's iss appears in the corresponding list of Entity Identifiers. Note that the list may be an empty list, which signifies that anyone can issue a Trust Mark with the trust_mark_type in question.

Saying here you MUST do this and somewhere else that you can also ignore this is not good. Either we add a section here or we also move this point somewhere else, IMO this would also fall under rolands point 3.

openid-federation-1_0.xml
```diff
@@ -3753,87 +3794,44 @@
...
+             <t>The instance MUST be a signed JWT</t>
+            <t>The instance MUST have a <spanx style="verb">typ</spanx> header with the value
+                <spanx style="verb">trust-mark+jwt</spanx>
+            </t>
+            <t>The instance MUST have an <spanx style="verb">alg</spanx> (algorithm) header parameter with a value
+                that is an acceptable JWS signing algorithm; it MUST NOT be none.
+            </t>
+            <t>The Entity Identifier of the Entity whose Entity Configuration contains the instance MUST
+                match the value of the claim <spanx style="verb">sub</spanx> in the Trust Mark</t>
```
selfissued:
```suggestion
                match the value of the claim <spanx style="verb">sub</spanx> in the Trust Mark.</t>
```

openid-federation-1_0.xml
```diff
@@ -3753,87 +3794,44 @@
...
+            <t>The current time MUST be after the time represented by the iat (issued at) Claim
+                (possibly allowing for some small leeway to account for clock skew).</t>
+            <t>The current time MUST be before the time represented by the exp (expiration) Claim
```
selfissued:
```suggestion
            <t>The current time MUST be before the time represented by the <spanx style="verb">exp</spanx> (expiration) Claim
```

openid-federation-1_0.xml
```diff
@@ -7729,7 +7726,45 @@ HTTP/1.1 302 Found
         between federations.
         </t>
     </section>
-
+    <section anchor="trust_mark_policy" title="Trust Mark policies">
+        <t>
+            When validating trust marks in an Entity Statement, it can be split into three parts.
+            <list style="hanging">
+                <t hangText="Validating trust marks in the context of validating an Entity Statement">
```
selfissued:
```suggestion
                <t hangText="Validating Trust Marks in the Context of Validating an Entity Statement">
```

openid-federation-1_0.xml
```diff
@@ -7729,7 +7726,45 @@ HTTP/1.1 302 Found
...
+            <vspace/>
+                    According to the text on Entity Statement Validation (TODO reference to be added) validating a
+                    Trust Mark is confined to validating the syntax of the claim value and verifying the
+                    <spanx style="verb">trust_mark_type</spanx> connection.
+                </t>
+                <t hangText="Validating a specific Trust Mark">
+                    <vspace/>
+                    This is what is described in section <xref target="trust-mark-validation"/>.
+                    In order to validate a Trust Mark the entity must find a trust chain for the Trust Mark Issuer to a
+                    Trust Anchor the entity trusts. This has nothing to do with which federation that will later
```
selfissued:
```suggestion
                    Trust Anchor the Entity trusts. This has nothing to do with which federation that will later
```

openid-federation-1_0.xml
```diff
@@ -7729,7 +7726,45 @@ HTTP/1.1 302 Found
...
+                    be used for the application protocol.
+                </t>
+                <t hangText="Deciding which Trust Marks to use">
+                    <vspace/>
+                    A federation may have a policy that states that only trust marks that
+                    match a certain criteria SHOULD be used.
```
selfissued:
```suggestion
                    match certain criteria SHOULD be used.
```

openid-federation-1_0.xml
```diff
@@ -7729,7 +7726,45 @@ HTTP/1.1 302 Found
...
+                    <vspace/>
+                    An example of such a criteria could be that the Trust Mark's
```
selfissued:
```suggestion
                    An example of such criteria could be that the Trust Mark's
```

openid-federation-1_0.xml
```diff
@@ -7729,7 +7726,45 @@ HTTP/1.1 302 Found
...
-
+    <section anchor="trust_mark_policy" title="Trust Mark policies">
+        <t>
+            When validating trust marks in an Entity Statement, it can be split into three parts.
+            <list style="hanging">
+                <t hangText="Validating trust marks in the context of validating an Entity Statement">
+            <vspace/>
+                    According to the text on Entity Statement Validation (TODO reference to be added) validating a
+                    Trust Mark is confined to validating the syntax of the claim value and verifying the
+                    <spanx style="verb">trust_mark_type</spanx> connection.
```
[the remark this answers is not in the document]
rohe: I reference this part of Entity Statement Validation (bullet point 19).

".. validate that there is a trust_mark_type member whose value matches the trust_mark_type Claim value in the Trust Mark JWT that is the value of the trust_mark member. .."

I could verbatim copy that if that would make it clearer.

zachmann (approved):

PR #282, opened 2025-11-10
Define peer_trust_chain to enable Federation Integrity and Metadata Integrity

### LINKED ISSUE #100: Trust Anchor Mix-Up (Federation Integrity Property)

selfissued:
Fixes #100

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -6506,6 +6506,19 @@ GET /authorize?
 
 ]]></artwork>
             </figure>
+	    <t>
+	      When the <spanx style="verb">trust_chain</spanx> header parameter
+	      is included,
+	      the <spanx style="verb">peer_trust_chain</spanx> header parameter
+	      MAY also be included to provide a Trust Chain between the OP
+	      and the Trust Anchor the RP selected.
+	      The Peer Trust Chain contains metadata and policy values
+	      that the RP has chosen for the OP to use during the registration.
+	      The Trust Anchors selected in both Trust Chains MUST be the same.
+	      Inclusion of both Trust Chains enables achieving
+	      the Federation Integrity and Metadata Integrity properties,
+	      as defined in <xref target="App-Fed-Linkage"/>.
+	    </t>
             </section>
 
             <section title="Processing the Authentication Request" anchor="AuthzRequestProcessing">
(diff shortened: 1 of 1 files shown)


rohe:
In section 12.2 Explicit Registration it is stated that an RP can submit its Entity Configuration or an entire Trust Chain.

If submitting the Trust Chain the Entity Statement representing the RPs Entity Configuration could contain the peer_trust_chain and that Trust Chain would then be matched (to verify that they both ends at the same TA) against the submitted Trust Chain.

On the other hand if the RP decides to submit the Entity Configuration there is no Trust Chain for the RP in the request so no matching could be preformed.

I think you there for would have to add trust_chain to the registration request claims with the proviso that it should only be added if the request contained the Entity Configuration and not the entire Trust Chain.

cicnavi (approved):

PR #299, opened 2025-11-26
Make two claim registration section references more specific at Nat's suggestion

selfissued:
For the "keys" and "trust_chain" claims.


selfissued:
See the rendered registrations at https://openid.github.io/federation/mbj-nat-claims.html#name-json-web-token-claims-regis .  The diffs aren't that helpful because the `trust_chain` registration was moved down to place it in ascending section number order.

PR #312, opened 2025-12-02
clarify resolve response expiration

### LINKED ISSUE #311: Resolver response expiration with trust marks.
The response of a resolve request has an expiration which MUST be equal to the expiration of the trust chain.
If the response also includes trust marks that would expire earlier, wouldn't it make sense that the resolve response also expires at that time?

zachmann:
This fixes #311

### DIFF
--- a/openid-federation-1_0.xml
+++ b/openid-federation-1_0.xml
@@ -78,7 +78,7 @@
       </address>
     </author>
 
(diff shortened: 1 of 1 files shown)


selfissued:
@zachmann please merge main into this PR and then add a history entry at the end of the -46 history entries saying that this fixes #311.  Thanks.

PR #318, opened 2025-12-20
Replaced the trust chain in exampled 6.

### LINKED ISSUE #317: Example Trust Chain Header does not begin with an Entity Configuration
Hi,

According to [Section 4.1](https://openid.github.io/federation/main.html#name-beginning-and-ending-trust-) the beginning of a trust chain is an entity configuration, which is an entity statement where the _iss_ and _sub_ are identical (See [Section 3.1](https://openid.github.io/federation/main.html#name-claims-that-must-or-may-app)). 

This is not the case for the first entry of the trust chain in the example of [Figure 6](https://openid.github.io/federation/main.html#name-example-jws-header-with-a-t) with this decoded payload:

{
    "valid": false,
    "payload": {
        "metadata": {
            "openid_credential_issuer": { ... },
            "federation_entity": { ... }
        },
        "authority_hints": [
            "https://intermediate.eidas.example.org"
        ],
        "jwks": { ... },
        "sub": "https://credential_issuer.example.org",
        "iss": "https://intermediate.eidas.example.org",
        "iat": 1758527818,
        "exp": 1758827818
    },
    "header": { ... }
}

rohe:
Fixes #317


rohe:
I will fix that after the holidays.

